PR #7, opened 2023-10-07
Chavic/fixture

chavic:
## Added  More Tests/Fixtures
Adding more tests to match the latest spec of Uniffi-rs v0.24 and make refactoring easier.
Because some features are implicitly covered by other tests, the total number of tests is reduced.
#### Next Steps
- [ ] Refactoring
- [ ] Implement futures


gnunicorn:
@chavic how come the test suite is still failing? Maybe we just want to disable broken tests for now so we can use the test suite from now on as an actual CI passing point?

chavic:
> @chavic how come the test suite is still failing? Maybe we just want to disable broken tests for now so we can use the test suite from now on as an actual CI passing point?

I looked at the errors for a few seconds... noticed issues with the tooling for example 'cargo fmt' wasn't found. Could work on fixing the CI.... it'll probably make things faster in the long run

PR #35, opened 2024-05-31
Backporting futures

gnunicorn:
from @chavic 's working branch into latest `main`:

- reactives the future tests
- copies over the futures implementation code
- for that already takes over the new uniffi_function generation code (but not yet the singleton)
- expands it over object methods
- fixes the hello-world test failure problem
- cleans up old, not necessary code and prints

Best viewed one commit at a time as these explain in the name what is even happening.


gtalha07 (approved):
Looks good! Although you probably wanna look at some test fails (but irrelevant to this PR)

PR #36, opened 2024-07-01
UniFFI Dart Binding Generator Refactor and Added Async Support

chavic:
Here's a summary overview of what's in this PR, some things were skipped because the codebase was in a satisfactory state, and it may not include other possible improvements (undone)... This should be enough to continue adding features

- [x] Adopted modular structure for primitives (separate files for different types)
- [x] Reorganized project structure for better maintainability
- [x] Integrated explicit async support (`async_poll`, `async_complete`, `async_free`)
- [x] Ensured compatibility with existing synchronous operations
- [x] Implemented separate files for boolean, string, and duration types
- [x] Refactored existing primitive types for consistency
- [x] Enhanced record implementation with read, write, and allocation size methods
- [x] Implemented comprehensive enum support, including complex variants with fields
- [ ] Updated existing enum usages to leverage new capabilities
- [x] Integrated `external_type_package_name` method
- [ ] Documented usage of external type handling
- [x] Combined detailed FFI type conversion with a simplified approach
- [x] Implemented unified FFI conversion strategy
- [x] Integrated `DurationCodeType`
- [ ] Ensured proper usage in relevant parts of the codebase
- [x] Implemented `LiftRetVal` struct for more precise lifting
- [x] Updated existing lifting operations to use new structure
- [x] Implemented refined `UniffiInternalError`
- [x] Updated error handling throughout the codebase
- [x] Merged best practices from both codebases
- [x] Implemented a more structured approach to code generation
- [x] Merged `TypeHelpersRenderer`, keeping useful methods from both codebases
- [ ] Ensured consistent usage of type helpers throughout the project
- [ ] Evaluated and combined the best parts of object handling
- [ ] Optimized critical paths for better performance
- [ ] Implemented benchmarks for key operations


src/gen-forgetten.rs
```diff
@@ -0,0 +1,328 @@
...
+//             other.cdylib_name.clone()
+//         } else {
+//             self.cdylib_name.clone()
+//         };
+//         Self {
+//             package_name,
+//             cdylib_name,
+//         }
+//     }
+// }
```
gnunicorn, line 54: Do we still need this? Is there a good reason to add this?
```suggestion
```

src/gen/functions.rs
```diff
@@ -58,9 +60,60 @@ use super::render::{ TypeHelperRenderer};
...
+//             final api = $api;
+//             $body
+//         }
+//     }
+// }
+
 pub fn generate_function(func: &Function, type_helper: &dyn TypeHelperRenderer) -> dart::Tokens {
     if func.takes_self() {} // TODO: Do something about this condition
     let args = quote!($(for arg in &func.arguments() => $(&arg.as_renderable().render_type(&arg.as_type(), type_helper)) $(DartCodeOracle::var_name(arg.name())),));
+    let args = quote!($(for arg in &func.arguments() => $(&arg.as_renderable().render_type(&arg.as_type(), type_helper)) $(DartCodeOracle::var_name(arg.name())),));
```
gnunicorn, line 115: why do it twice?

src/gen/objects.rs
```diff
@@ -68,6 +69,8 @@ pub fn generate_object(obj: &Object, type_helper: &dyn TypeHelperRenderer) -> da
                 return $(cls_name)._(ptr);
             }
 
+            // TODO: Bring back object lowering from main
```
gnunicorn, line 69: is this still accurate?

src/gen/primitives/string.rs
```diff
@@ -34,8 +33,7 @@ impl Renderable for StringCodeType {
                 }
 
                 static int allocationSize([String value = ""]) {
-                    // FIXME: doing this twice for every string is bad
-                    return utf8.encoder.convert(value).length + 4; // Four additional bytes for the length data
+                    return utf8.encoder.convert(value).length + 4;
```
gnunicorn, line 36: I'd prefer we kept the code comment so people reading it understand what that number is about:
```suggestion
                    return utf8.encoder.convert(value).length + 4; // Four additional bytes for the length data
```

src/gen/primitives/string.rs
```diff
@@ -34,8 +33,7 @@ impl Renderable for StringCodeType {
                 }
 
                 static int allocationSize([String value = ""]) {
-                    // FIXME: doing this twice for every string is bad
```
gnunicorn, line 37: technically this fixme hasn't been sold, has it?!? ... but well, I am not a fan of using fixme and other flags in the code so I am okay with removing it.

src/gen/types.rs
```diff
@@ -575,8 +575,14 @@ impl Renderer<(FunctionDefinition, dart::Tokens)> for TypeHelpersRenderer<'_> {
...
-                Uint8List asUint8List() {
-                    return data.cast<Uint8>().asTypedList(len);
+                RustBuffer reserve(int additionalCapacity) {
+                return rustCall((status) => $(DartCodeOracle::find_lib_instance()).$(self.ci.ffi_rustbuffer_reserve().name())(this, additionalCapacity, status));
+                }
+
+                Uint8List asTypedList() {
+                final dataList = data.asTypedList(len);
+                final byteData = ByteData.sublistView(dataList);
+                return Uint8List.view(byteData.buffer);
```
gnunicorn: what's the benefit of this over the one-liner this replaces?
```
                    return data.cast<Uint8>().asTypedList(len);
```

src/gen/enums.rs
```diff
@@ -89,34 +86,32 @@ pub fn generate_enum(obj: &Enum, type_helper: &dyn TypeHelperRenderer) -> dart::
...
                         $(for f in obj.fields() =>
-                            final $(var_name(f.name()))_lifted = $(f.as_type().as_codetype().ffi_converter_name()).read(api, Uint8List.view(buf.buffer, new_offset));
+                            final $(var_name(f.name()))_lifted = $(f.as_type().as_codetype().ffi_converter_name()).read(Uint8List.view(buf.buffer, new_offset));
                             final $(var_name(f.name())) = $(var_name(f.name()))_lifted.value;
                             new_offset += $(var_name(f.name()))_lifted.bytesRead;
                         )
                         return LiftRetVal($(class_name(obj.name()))$cls_name._(
                             $(for f in obj.fields() => $(var_name(f.name())),)
-                        ), new_offset);
+                        ), new_offset - buf.offsetInBytes);
```
gnunicorn, line 107: can you elaborate on this change? It seems odd to wrong ... considering that we start out with `int new_offset = buf.offsetInBytes;` in L98, why would we add stuff and then substract that same offset again? Did you change the meaning of the return value?

src/gen/functions.rs
```diff
@@ -1,59 +1,150 @@
...
-    let inner = quote! {
-    rustCall(api, (res) =>
-        _$(&fn_name)(
-            $(for arg in &fun.arguments() => $(DartCodeOracle::type_lower_fn(&arg.as_type(), quote!($(var_name(arg.name()))))),)
-        res)
-    )
-    };
 
-    let (ret, body) = if let Some(ret) = fun.return_type() {
+// #[allow(unused_variables)]
```
gnunicorn, line 12: What speaks against just removing this? We can still recover it from the git history if we really need to (but in my experience this is very unlikely to happen ;) ).

src/gen/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+//             other.cdylib_name.clone()
+//         } else {
+//             self.cdylib_name.clone()
+//         };
+//         Self {
+//             package_name,
+//             cdylib_name,
+//         }
+//     }
+// }
```
gnunicorn, line 51:
```suggestion

```

src/gen/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+}
+
+impl Config {
+    pub fn package_name(&self) -> String {
+        if let Some(package_name) = &self.package_name {
+            package_name.clone()
+        } else {
+            "uniffi".into()
+        }
```
gnunicorn, line 69: As I am just seeing it:
```suggestion
        self.package_name.clone().unwrap_or_else(||"uniffi".into())
```

src/gen/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    pub fn cdylib_name(&self) -> String {
+        if let Some(cdylib_name) = &self.cdylib_name {
+            cdylib_name.clone()
+        } else {
+            "uniffi".into()
+        }
+    }
```
gnunicorn, line 78:
```suggestion
        self.cdylib_name.clone().unwrap_or_else(|| "uniffi".into())
```

src/gen/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            }
+
+            void initialize() {
+                _UniffiLib._open();
+            }
+
+            void ensureInitialized() {
+                _UniffiLib._checkApiVersion();
+                _UniffiLib._checkApiChecksums();
+            }
```
gnunicorn, line 271: why two functions? Let's wrap it into one.
```suggestion
            void initialize() {
                _UniffiLib._open();
                _UniffiLib._checkApiVersion();
                _UniffiLib._checkApiChecksums();
            }
```

src/gen/oracle.rs
```diff
@@ -62,47 +61,56 @@ impl DartCodeOracle {
...
         match name.strip_suffix("Error") {
             None => name,
             Some(stripped) => format!("{stripped}Exception"),
         }
     }
 
+    pub fn find_lib_instance() -> dart::Tokens {
+        quote!(_UniffiLib.instance)
+    }
+
```
gnunicorn, line 81: as we hard-code this in the builder anyways, why make it a function on the Oracle and not just use `_UniffiLib.instance` everywhere instead?

src/gen/primitives/macros.rs
```diff
@@ -40,25 +61,32 @@ macro_rules! impl_renderable_for_primitive {
 
                 quote! {
                     class $cl_name {
-                        static $type_signature lift(Api api, RustBuffer buf) {
-                            return $cl_name.read(api, buf.asUint8List()).value;
-                        }
-                        static LiftRetVal<$type_signature> read(Api api, Uint8List buf) {
+                        // static $type_signature lift($type_signature value) {
+                        //     return $cl_name.read(buf.asUint8List()).value;
+                        // }
```
gnunicorn, line 66:
```suggestion
```

src/gen/primitives/macros.rs
```diff
@@ -40,25 +61,32 @@ macro_rules! impl_renderable_for_primitive {
...
-                            byteData.set$conversion_name(0, value$endian);
-                            return toRustBuffer(api, Uint8List.fromList(buf.toList()));
-                        }
+                        // static RustBuffer lower($type_signature value) {
+                        //     final buf = Uint8List($cl_name.allocationSize(value));
+                        //     final byteData = ByteData.sublistView(buf);
+                        //     byteData.set$conversion_name(0, value$endian);
+                        //     return toRustBuffer(Uint8List.fromList(buf.toList()));
+                        // }
+
```
gnunicorn, line 81:
```suggestion
```

src/gen-forgetten.rs
gnunicorn, line 1: I understand this filename. what is `forgetten` supposed to convey?

gnunicorn: and what is this file in the first place? It looks like the same as `gen/mod.rs` ... is is a broken move? I think this file should be gone ...

src/gen/primitives/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+            | Type::Float32
+            | Type::Float64
+            | Type::Duration => num_str,
+            _ => panic!("Unexpected literal: {} is not a number", num_str),
+        }
+    }
+
+    match literal {
+        Literal::Boolean(v) => format!("{}", v),
+        Literal::String(s) => format!("\"{}\"", s),
```
gnunicorn, line 36: Just noticed. I think dart has a preference for `'`:
 ```suggestion
        Literal::String(s) => format!("'{}'", s),
```

File: src/gen/primitives/mod.rs [hunk 17]
```diff
@@ -0,0 +1,171 @@
...
+impl_renderable_for_primitive!(Int32CodeType, "int", "Int32", 4);
+impl_renderable_for_primitive!(Int64CodeType, "int", "Int64", 8);
+impl_renderable_for_primitive!(UInt8CodeType, "int", "UInt8", 1);
+impl_renderable_for_primitive!(UInt16CodeType, "int", "UInt16", 2);
+impl_renderable_for_primitive!(UInt32CodeType, "int", "UInt32", 4);
+impl_renderable_for_primitive!(UInt64CodeType, "int", "UInt64", 8);
+impl_renderable_for_primitive!(Float32CodeType, "double", "Double32", 4);
+impl_renderable_for_primitive!(Float64CodeType, "double", "Double64", 8);
+
+// pub fn generate_wrapper_lifters() -> dart::Tokens {
```
gnunicorn, line 83: code hygiene ... let's just remove this.

src/gen/records.rs
```diff
@@ -80,21 +75,26 @@ pub fn generate_record(obj: &Record, type_helper: &dyn TypeHelperRenderer) -> da
                 ), new_offset);
             }
 
-            static RustBuffer lower(Api api, $cls_name value) {
+            static RustBuffer lower( $cls_name value) {
                 final total_length = $(for f in obj.fields() => $(f.as_type().as_codetype().ffi_converter_name()).allocationSize(value.$(var_name(f.name()))) + ) 0;
                 final buf = Uint8List(total_length);
-                $ffi_conv_name.write(api, value, buf);
-                return toRustBuffer(api, buf);
+                $ffi_conv_name.write(value, buf);
```
gnunicorn, line 81: to keep it consistent?
```suggestion
                write(value, buf);
```

src/gen/render/mod.rs
```diff
@@ -112,34 +74,20 @@ impl<T: AsType> AsRenderable for T {
             Type::String => Box::new(primitives::StringCodeType),
             Type::Duration => Box::new(primitives::DurationCodeType),
             Type::Object { name, .. } => Box::new(objects::ObjectCodeType::new(name)),
-            Type::Optional { inner_type, .. } => Box::new(compounds::OptionalCodeType::new(
```
gnunicorn, line 115: aren't these necessary for the latest uniffi dep?

gnunicorn (approved):
Looking good. Just a few hygiene and a bunch of minor inquiries ... and one change I don't understand and need some explanation for ...

src/gen/functions.rs
chavic: this was a leak during the merging

src/gen/objects.rs
chavic: Yeah, I noticed a missing lower function for objects... It may be in the backup, just needed to remind myself

src/gen/primitives/string.rs
chavic: yeah, I will bring it back

src/gen/primitives/string.rs
chavic: okay, but probably not too bad of an idea to bring it back

src/gen-forgetten.rs
chavic: I mainly kept it for comparison purposes, it leaked in through a different commit

PR #61, opened 2025-07-15
Adding Comprehesive Test Coverage

chavic:
This is a draft PR to contain all the current work on adding the full set of fixtures/tests


chavic:
Yeah... proc macro support isn't supported yet, there's more than one error in there, they're supposed to fail right now... usually, they'd be added to the exclude list so they don't mess with the CI since we know we don't have support. I'll do that later, right now, there are still about 10-15 fixtures left to add

chavic:
Fixed a bunch of CI issues. From now on, a fixture will be added to the workspace when a work-specific fixture starts, making it clear what has been done and what’s pending. References are also included in the TODO.md

.github/workflows/tests.yml
```diff
@@ -23,10 +23,12 @@ jobs:
...
-      - uses: FrancisRussell/ferrous-actions@v0.1.0-beta.1
-        name: Cargo Cache
-        with:
-          command: cache
+      # TODO: Re-enable when GitHub cache service is stable
+      # Temporarily disabled due to "Cache service responded with 400" errors
+      # - uses: FrancisRussell/ferrous-actions@v0.1.0-beta.1
+      #   name: Cargo Cache
+      #   with:
+      #     command: cache
```
spacebear21, line 31: In rust-payjoin we use https://github.com/Swatinem/rust-cache for caching, it's actively maintained so probably uniffi-dart should move awayt from ferrous-actions which hasn't seen activity in 2 years? See https://github.com/payjoin/rust-payjoin/blob/master/.github/workflows/rust.yml

spacebear21 (approved):
Great to have all these fixtures ready for test-driven development of the remaining features! Thanks also for fixing the CI.

In the future it will be easier for review to ensure that each commit follows the formatting/linting rules instead of separating functional vs. formatting changes. This can be enforced with e.g. format-on-save in editor, or pre-commit git hooks.

PR #82, opened 2025-08-24
Default to temporary directories with configurable output options

chavic:
- Use camino-tempfile::tempdir by default with automatic cleanup
- Add TestConfig for full control over test execution
- Support env overrides: UNIFFI_DART_TEST_DIR, UNIFFI_DART_NO_DELETE, UNIFFI_DART_FAILURE_DELAY
- Resolve relative custom dirs against workspace root
- Preserve temp files when no-delete is set (uses std::mem::forget)
- Remove .tmp_tests references and folder - temp dirs are cleaner by default

Backwards compatible: existing run_test() continues to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,5 @@ og-fixtures/*
 rust-toolchain.toml
 internal_todo.md
 ISSUES.md
-.tmp_tests/
 convert_fixtures.sh
 .DS_Store
--- a/src/testing.rs
+++ b/src/testing.rs
@@ -1,6 +1,7 @@
 use crate::gen;
 use anyhow::{bail, Result};
 use camino::{Utf8Path, Utf8PathBuf};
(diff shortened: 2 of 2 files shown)


spacebear21:
Since this PR does not depend on Hashmap changes, it should just contain the tempfile commit (dd1d75ce0bd58952bb0df5773193e0cd7b48f58e) and the first commit can be dropped. This enables PRs to be reviewed and merged independently from each other.

chavic:
> There are now even more unrelated commits then before? To be clear this PR should just contain [792ea82](https://github.com/Uniffi-Dart/uniffi-dart/commit/792ea82fb1b3a37e1cbce45b0ac36e30482eab54)

I'm not sure if the merge conflict had anything to do with it... I'll try again 🤔, though I remember fixing it...

PR #99, opened 2025-11-22
Shift to Native Assets + update to uniffi 0.30.0

kumulynja:
This PR updates uniffi to 0.30.0 and with some other fixes I could make binding generation work with it for [bdk-ffi](https://github.com/bitcoindevkit/bdk-dart/pull/12) which uses HashMap/Map and pro-macros, so possibly 2 of the 5 critical issues are solved with them as well, but I should probably add specific tests for them still.

A second pretty drastic change done here was moving from path based opening of the native libs to using Native Assets, which recently became a stable feature in both Dart and Flutter and is the recommended way now to use native binaries.

### DIFF
--- a/src/gen/mod.rs
+++ b/src/gen/mod.rs
@@ -139,7 +139,8 @@ impl<'a> DartWrapper<'a> {
 
                     for arg in fun.arguments() {
                         let arg_name = arg.name();
-                        let native_type = DartCodeOracle::ffi_native_type_label(Some(&arg.type_()), ci);
+                        let native_type =
+                            DartCodeOracle::ffi_native_type_label(Some(&arg.type_()), ci);
                         let dart_type = DartCodeOracle::ffi_dart_type_label(Some(&arg.type_()), ci);
 
                         native_arg_vec.push(native_type);
@@ -171,7 +172,7 @@ impl<'a> DartWrapper<'a> {
             definitions
         }
 
-        let asset_id_suffix = &self.config.asset_id();  // e.g., "uniffi:hello_world"
+        let asset_id_suffix = &self.config.asset_id(); // e.g., "uniffi:hello_world"
 
         quote! {
             library $package_name;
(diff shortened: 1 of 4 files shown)


spacebear21:
Nice work! Glad uniffi-dart is working for you. For payjoin-ffi we are stuck on uniffi 0.29 for the time being (see https://github.com/payjoin/rust-payjoin/issues/1167) so it would be nice to separate the code fixes from the uniffi upgrade, such that we can maintain two branches to support 0.30 and keep supporting 0.29 for some time.

kumulynja:
@spacebear21 Since I was a bit stuck with how to bundle/package binaries to publish packages with generated bindings on pub.dev or easily use the packages from other projects, I made quite radical changes, but I think they are worth it:
This PR refactors uniffi-dart to use Native Assets, the recent and recommended way to use native libraries in Dart/Flutter.

The main advantage it has is that the native lib shouldn't be pre-compiled somewhere or cargo-kit shouldn't be used to hook into the Flutter build process anymore. With Native Assets and [native_toolchain_rs](https://pub.dev/packages/native_toolchain_rs) a build hook does all of this for the correct target automatically.
This makes it also possible to use the bindings for pure Dart CLI's, server side dart and Flutter apps (Desktop/iOS/Android) all without any custom configurations per platform.

I think this is the way to go, but of course this needs more eyes and thorough testing. As of now it was tested only very limited with bdk-dart here: https://github.com/bitcoindevkit/bdk-dart/pull/12
As you can see, no special config is needed for neither ios or the example apps. The build hook does the magic dynamically.

Let me know what you think

spacebear21:
The downstream bindings test is expected to fail because it uses the master branch from rust-payjoin. There is a PR to add native 0.30 + native asset support (https://github.com/payjoin/rust-payjoin/pull/1208) but it depends on this PR so there is a chicken-and-egg problem. I've tested rust-payjoin locally with the changes in this PR and am confident it works like a charm, so merging this now.

Issue #66, opened 2025-08-07
Complete Async/Await Support (Futures)


chavic:
**Problem:**
Basic async infrastructure exists but needs completion. Current implementation fails due to trait compatibility issues.

**Affected Fixture:**

- `dart_async` - Fails with "trait `SayAfterUdlTrait` is not dyn compatible"

**Acceptance Criteria:**

- [ ] Fix dyn compatibility issues with async traits
- [ ] Support `async fn` in trait definitions
- [ ] Generate proper Dart `Future<T>` types  
- [ ] Handle async error propagation
- [ ] Support async callbacks
- [ ] `dart_async` fixture passes all tests

**Technical Implementation:**

- Resolve E0038 dyn compatibility errors
- Update trait object generation for async methods
- Ensure proper Future handling in Dart

spacebear21:
Closed by #102

PR #102, opened 2025-11-27
Add async constructor & callback support for Dart bindings

chavic:
Finally taught the Dart bindings how to handle async functions. The runtime now keeps a handle map for Rust futures and foreign (Dart) futures; we generate static factory methods for async constructors, and callback interfaces can actually await their Dart implementations. I also cleaned up the dart_async fixture tests so they call those factories instead of pretending constructors can be async. This should address #66 

**Summary**
- overhaul `uniffiRustCallAsync` and add shared `UniffiForeignFuture` runtime glue (handle map, cancellation)
- generate async factories for object constructors and route async free functions through the new helper
- teach callback codegen to emit foreign future structs/typedefs and drive async Dart callbacks safely
- update `dart_async` fixture/tests to call the async factories (`Megaphone.new_()`, `UdlMegaphone.new_()`, etc.) and adjust timing expectations

**Testing**
```bash
cargo test -p dart_async
```

### DIFF
--- a/src/gen/callback_interface.rs
+++ b/src/gen/callback_interface.rs
@@ -1,7 +1,6 @@
 use crate::gen::CodeType;
 use genco::prelude::*;
 use heck::ToUpperCamelCase;
-use std::collections::BTreeSet;
 use uniffi_bindgen::interface::Type;
 use uniffi_bindgen::interface::{
     ffi::{FfiStruct, FfiType},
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,6 @@
+# Configure extra Mach-O header padding on macOS so that
+# Dart's bundler can safely rewrite install names for native assets.
+[target.'cfg(target_os = "macos")']
+rustflags = ["-C", "link-args=-Wl,-headerpad_max_install_names"]
+
+
```
spacebear21: nit: trailing newlines

spacebear21 (approved):
tACK 6558d4744e825a32a857a75fdfbf07a8c7f375b2, thanks for the prompt fix!

PR #103, opened 2025-12-29
Fix/callback interface code generation

kumulynja:
I encountered some errors when generating bindings for a UDL with `callback interface`s that return or accept primitive types (`integers`, `floats`).

It was caused by the following things that got fixed in this PR:

1. `callback_out_return_type` function only handled `Boolean` and `Object` types explicitly, causing all primitive integers and floats to fall through to the catch-all case which returned `Pointer<RustBuffer>` instead of the correct `Pointer<Uint64>`, `Pointer<Int32>`, etc.

2. `callback_return_handling` function used `outReturn.ref =` for all types, but primitive types require `outReturn.value =` since they're stored directly, not as references.

3. Constructor argument lowering used `type_lower_fn` directly, which returns `Pointer<Void>` for callback interfaces. FFI functions expect `int` (the address), requiring `.address` to be appended.

### DIFF
--- a/src/gen/oracle.rs
+++ b/src/gen/oracle.rs
@@ -619,8 +619,8 @@ impl DartCodeOracle {
             Type::Object {
                 imp: ObjectImpl::CallbackTrait,
                 ..
-            }
-            | Type::CallbackInterface { .. } => quote!($base_lower.address),
+            } => base_lower,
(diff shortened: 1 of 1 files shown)


spacebear21 (approved):
ACK 2c7e71ef900249d2e4b2e45d454d3d1aeba621b3

Issue #107, opened 2026-02-12
UnexpectedEnumCase in decoding


reez:
A bdk-dart user reported a crash when calling wallet.transactions() from Dart

This occurs during transaction decoding in generated Dart bindings (CanonicalTx / ChainPosition path). It appears to happen while reading sequences of transactions, not when invoking the call itself.

I opened up an issue on bdk-dart https://github.com/bitcoindevkit/bdk-dart/issues/46

Opening up an issue here too because my initial guess is that the issue originates in uniffi-dart because the failure occurs in generated Dart FFI decoding code?

spacebear21:
Merged #108 , let's have the user confirm it fixes both bugs mentioned above before closing this issue

PR #108, opened 2026-02-12
Fix UnexpectedEnumCase in decoding

reez:
Looking to address #107  UnexpectedEnumCase / decode errors for non-flat enums in Dart bindings when enums are nested in records/sequences.  

I think root cause was offset accounting where generated non-flat enum converters were returning absolute offsets from read+write, while parent converters consume relative byte counts.

Changes made in PR to normalize non-flat enum read+write bytes to relative offsets and adding regression coverage, but feel free to let me know if I'm on the wrong track or missing something.

## More detail

The enum in bdk-dart that caused the issue was:

```
enum ChainPosition {
  Confirmed { confirmation_block_time: ..., transitively: ... },
  Unconfirmed { timestamp: ... },
}
```

bdk-dart has other non-flat enums but i think this showed up as an issue specifically because wallet.transactions() path decodes a sequence of records contaning ChainPosition

### DIFF
--- a/fixtures/dart_async/src/lib.rs
+++ b/fixtures/dart_async/src/lib.rs
@@ -176,9 +176,37 @@ pub struct MyRecord {
     pub b: u32,
 }
 
+#[derive(uniffi::Enum)]
+pub enum AsyncItemState {
+    Ready { timestamp_ms: u64 },
+    Pending { reason: String },
+}
+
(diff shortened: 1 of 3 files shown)


reez:
User hit another bug that I think is related to this, both listed on #107

PR #111, opened 2026-02-19
Downstream test bdkdart

spacebear21:
uniffi-dart is growing fast, as [bdkdart in Bull Bitcoin Mobile](https://github.com/SatoshiPortal/bullbitcoin-mobile/pull/1874) is looking to be the first production application!

This PR adds bdkdart as another downstream bindings library to run regression tests on in CI.


reez (approved):
ACK c4b65ca86c6baa7878efa601acd6f1865db8456d

This makes sense to me, good idea and good implementation from my pov

PR #121, opened 2026-03-29
use Optional FfiConverter's lower() for inner types

kumulynja:
The old non-String case was manually prepending a discriminant byte on top of a RustBuffer that already contained one.
This caused uniffi conversion errors in generated bindings.


chavic (approved):
I enable the tests that would have caught this

PR #125, opened 2026-04-20
feat: uniffi 0.31

Psycarlo:


Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "Apache-2 or MIT"
 homepage = "https://github.com/acterglobal/uniffi-dart"
 description = "Dart Frontend for UniFFI"
-rust-version = "1.85"
+rust-version = "1.87"
```
spacebear21: Is the MSRV bump strictly necessary to support uniffi 0.31? Based on their changelog the MSRV [hasn't changed since 0.29](https://github.com/mozilla/uniffi-rs/blob/main/CHANGELOG.md#v0291-backend-crates-v0291---2025-03-18) though I'm not sure if they pay close attention to this, seems like more of a policy

Psycarlo: Thanks for the review. I'll take a look and update with the necessary changes.

Psycarlo:
@spacebear21 Reverted to `1.85`. What else do we need for `uniffi 0.31`?

spacebear21:
> What else do we need for uniffi 0.31?

I haven't tested myself but everything is green so I'm ok to merge it if it's working for you!

The only thing I'd like to do is first https://github.com/Uniffi-Dart/uniffi-dart/pull/127 and cut a bugfix release targeting 0.30.0 before we merge this to main, then will make a release branch for 0.31.x so those two targets can coexist, as we're not ready to upgrade to 0.31 yet at this time in payjoin-ffi.

PR #126, opened 2026-04-20
Add WASM web target support to uniffi-dart bindings

### LINKED ISSUE #106: Add Web Target (WASM) Support
Currently, `uniffi-dart` generates bindings based on `dart:ffi`, which restricts usage to native platforms (Android, iOS, macOS, Windows, Linux). To support the web platform, Dart requires interoperability with WebAssembly (WASM) via `dart:js_interop` and `package:web`, as `dart:ffi` is not available in browsers.

I propose a dual strategy that allows a single generated Dart package to support both native and web platforms through conditional imports.

1. WASM-bindgen Integration
The Rust crates will need to be compiled to `wasm32-unknown-unknown`. We will leverage `wasm-bindgen` to generate the JavaScript glue code required to load the WASM module. The generated Dart code will interact with this JavaScript layer.

2. Conditional Compilation in Dart
The generated Dart code will be split into three parts to handle platform differences:

* lib.dart (Entry Point):
`export 'src/lib_native.dart' if (dart.library.js_interop) 'src/lib_web.dart';`
* src/lib_native.dart (Existing FFI)
* src/lib_web.dart (New JS Interop): Implements the same API surface but using `dart:js_interop`, Maps Rust types to JS types (e.g., `RustBuffer` -> `JSUint8Array`), Loads the WASM module via `package:web` / JS global context.

Origin: [issue 40 bdk-dart](https://github.com/bitcoindevkit/bdk-dart/issues/40) as per user [request](https://github.com/bitcoindevkit/bdk-dart/issues/16#issuecomment-3745118270)

Johnosezele:
This PR resolves [#106](https://github.com/Uniffi-Dart/uniffi-dart/issues/106).

Notes to reviewers:
In Scope:
- Sync functions (all types: primitives, strings, bytes, records, enums, optionals, sequences, maps)
- Objects (constructors, methods, explicit `dispose()`, best-effort `FinalizationRegistry` fallback)
- Async functions (Rust futures exposed as `JSPromise` via `wasm_bindgen_futures`, converted to Dart `Future` via `.toDart`)
- Both UDL and proc-macro input modes (the shim wraps C-ABI symbols, which are identical regardless of input mode)

Out of this pr scope, until this effort is proven:
- Callback interfaces
- Callback trait objects
- Async callbacks
- Streams

New features

- [ ] Dual-target binding layout via conditional exports (`*_native.dart`, `*_web.dart`, `*_stub.dart`)
- [ ] Stub generation that preserves public API shapes but throws `UnsupportedError` on unsupported platforms

### DIFF
--- a/src/gen/web/functions.rs
+++ b/src/gen/web/functions.rs
@@ -1,0 +1,126 @@
+use genco::prelude::*;
+use heck::ToUpperCamelCase;
+use uniffi_bindgen::interface::{AsType, Function, Type};
+
+use crate::gen::oracle::DartCodeOracle;
+
+use super::oracle::WebCodeOracle;
+
+pub fn is_supported_function(func: &Function) -> bool {
+    !func.is_async()
+        && func.throws_type().is_none()
+        && func
+            .arguments()
+            .iter()
+            .all(|arg| is_supported_type(&arg.as_type()))
+        && func.return_type().map(is_supported_type).unwrap_or(true)
+}
+
+pub fn is_supported_type(type_: &Type) -> bool {
+    WebCodeOracle::dart_type_label(type_).is_some() && WebCodeOracle::js_type_label(type_).is_some()
+}
+
+pub fn uses_bytes(func: &Function) -> bool {
+    func.arguments()
+        .iter()
+        .any(|arg| matches!(arg.as_type(), Type::Bytes))
+        || func
+            .return_type()
+            .map(|ret| matches!(ret, Type::Bytes))
+            .unwrap_or(false)
+}
+
+pub fn generate_function(func: &Function, module_name: &str, namespace: &str) -> dart::Tokens {
+    let public_name = DartCodeOracle::fn_name(func.name());
+    let external_name = format!("_uniffiWeb{}", func.name().to_upper_camel_case());
+    let external_call_name = external_name.clone();
+    let js_export_name = format!(
+        "{}_{}",
+        namespace.replace('-', "_"),
+        func.name().replace('-', "_")
+    );
+
+    let public_args: Vec<dart::Tokens> = func
+        .arguments()
+        .iter()
+        .map(|arg| {
+            let arg_name = DartCodeOracle::var_name(arg.name());
+            let arg_type = WebCodeOracle::dart_type_label(&arg.as_type())
+                .expect("supported function argument should have Dart type");
+            quote!(required $arg_type $arg_name)
+        })
+        .collect();
+
+    let external_args: Vec<dart::Tokens> = func
+        .arguments()
+        .iter()
+        .map(|arg| {
+            let arg_name = DartCodeOracle::var_name(arg.name());
+            let arg_type = WebCodeOracle::js_type_label(&arg.as_type())
+                .expect("supported function argument should have JS type");
+            quote!($arg_type $arg_name)
+        })
+        .collect();
+
+    let call_args: Vec<dart::Tokens> = func
+        .arguments()
+        .iter()
+        .map(|arg| {
+            let arg_name = DartCodeOracle::var_name(arg.name());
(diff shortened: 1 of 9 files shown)


chavic:
@Johnosezele nice work, will keep an eye on this...

chavic:
@Johnosezele, what do you think about pulling 434099f7a06409cf1c934e4815426c8f19382649 into its own PR?

Johnosezele:
>what do you think about pulling https://github.com/Uniffi-Dart/uniffi-dart/commit/434099f7a06409cf1c934e4815426c8f19382649 into its own PR?

Originally, since I added `Type::Timestamp => DateTime` in `src/gen/types.rs` for c3a9934ff754f66014278db7d1c3d311a94997fc, I needed to add test coverage for it...
So the reason I aded time-types was to put a fixture that forces timestamp shaped generated bindings to be built and tested, so a future refactor can’t silently re-break timestamp handling....

in the process of impl I learned that it does more than validate the stub path it also exercises older native generator paths that still do not support `Type::Timestamp` so adding time-types exposed a broader pre-existing generator gap... hence 434099f7a06409cf1c934e4815426c8f19382649

To ease review, I shall move 434099f7a06409cf1c934e4815426c8f19382649 into its own PR

src/gen/web/functions.rs
```diff
@@ -0,0 +1,126 @@
...
+                    $(&external_call_name)($call_params);
+                } catch (error) {
+                    _throwWebError(error);
+                }
+            },
+        )
+    };
+
+    quote! {
+        @JS($(quoted(format!("{module_name}.{js_export_name}"))))
```
reez, line 119: Where is the JS shim is created?

src/gen/web/types.rs
```diff
@@ -0,0 +1,97 @@
...
+pub fn generate_web_runtime(module_name: &str) -> dart::Tokens {
+    quote! {
+        @JS($(quoted(format!("{module_name}.init"))))
+        external JSPromise<JSAny?> _uniffiWasmInit(JSString wasmUrl);
+
+        Future<void>? _initFuture;
+        bool _initialized = false;
+        String? _initializedWasmPath;
+
+        Future<void> ensureInitialized({String? wasmPath}) {
```
reez, line 12: im seeing web export with `Future<void> ensureInitialized({String? wasmPath})` and native stub with `void ensureInitialized()`, should they match to keep same public api shape?

src/gen/web/mod.rs
```diff
@@ -0,0 +1,174 @@
...
+
+pub fn unsupported_web_api_names(ci: &ComponentInterface) -> Vec<String> {
+    let mut unsupported: Vec<String> = ci
+        .function_definitions()
+        .iter()
+        .filter(|func| !functions::is_supported_function(func))
+        .map(|func| format!("function {}", func.name()))
+        .collect();
+
+    unsupported.extend(
```
reez, line 128: This adds every object definition to the unsupported web API list, Is object support still intended to be in scope for this PR, or should the PR description be narrowed to the function only scaffold that’s implemented here? I could totally be missing something though

reez (commented):
Awesome PR to open! Added a couple comments/questions

src/gen/web/functions.rs
Johnosezele: it's not landed just yet, these `@JS(...)` declarations are defining the interface that later commit will provide.

src/gen/web/types.rs
Johnosezele: yes, they should match in the final public API...the mismatch here is intentional for now.

src/gen/web/mod.rs
Johnosezele: object support is still intended. The reson objects are listed as unsuported here is to make `web_unsupported_policy = "error"` correctly reject unsuported exported surfaces instead of silently succeding...

PR #134, opened 2026-04-30
fix: make generated dart helper ordering deterministic

### LINKED ISSUE #133: Dart codegen emits FFI helper converters in nondeterministic order
`uniffi-dart` can generate equivalent Dart bindings with different `FfiConverter*` helper declaration order across repeated runs.

Observed from bitcoindevkit/bdk-dart#68: repeated generation kept the same generated converter class set, but reordered helper declarations in `lib/bdk.dart`, creating large noisy diffs.

Expected: repeated codegen with the same inputs should produce byte-stable Dart output.

Note: this does not appear to be the same pattern used by Mozilla’s built-in UniFFI generators. Swift/Kotlin/Ruby render helpers by `ci.iter_local_types()`, which comes from UniFFI’s ordered `TypeUniverse` (`BTreeSet`). This looks specific to `uniffi-dart`’s helper include tracking.

Likely cause?: Dart helper include tracking uses unordered Rust collections before rendering helper definitions.

reez:
## Summary

Fixes #133 

Makes Dart helper converter generation deterministic by storing include once helper names in ordered collections.

Previously, `TypeHelpersRenderer` tracked helper types with `HashMap`/`HashSet` and rendered by iterating that map. Since Rust hash collection iteration order is not stable between processes, repeated generation could reorder `FfiConverter*` helper declarations and create noisy generated diffs.

This switches the helper registry to `BTreeMap`/`BTreeSet` and adds a regression test for sorted helper name iteration.

I also sanity checked Mozilla’s built in UniFFI generators. Swift/Kotlin/Ruby render helper definitions via `ci.iter_local_types()`, which comes from UniFFI’s ordered `TypeUniverse` (`BTreeSet`). So this change aligns Dart helper rendering with the deterministic ordering pattern used by the upstream generators rather than preserving plugin specific hash map iteration.

### DIFF
--- a/src/gen/types.rs
+++ b/src/gen/types.rs
@@ -1,5 +1,5 @@
-use std::collections::{BTreeSet, HashSet};
-use std::{cell::RefCell, collections::HashMap};
+use std::cell::RefCell;
+use std::collections::{BTreeMap, BTreeSet};
 
 use genco::prelude::*;
 use uniffi_bindgen::interface::AsType;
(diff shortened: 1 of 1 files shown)


spacebear21 (approved):
utACK, thanks @reez !

PR #135, opened 2026-05-07
Support Rust-backed with_foreign callback traits

chavic:
This fixes Dart bindings for `#[uniffi::export(with_foreign)]` callback traits when Rust returns an `Arc<dyn Trait>`.

UniFFI uses tagged callback/interface handles: odd values are foreign objects stored in the foreign-language handle map, while even values are Rust-owned object pointers. The Dart generator previously treated callback interface handles as handle-map entries only, which caused Rust-owned callback trait objects to fail with stale-handle errors.

This enables callback-heavy fixture coverage, adds regressions for the Payjoin-shaped `with_foreign` callback trait case, teaches callback interface lifting/lowering to distinguish tagged handles, and generates Dart proxy classes for Rust-backed callback traits so they can be called from Dart and passed back into Rust.

### DIFF
--- a/fixtures/callbacks/test/callbacks_test.dart
+++ b/fixtures/callbacks/test/callbacks_test.dart
@@ -173,6 +173,25 @@ void main() {
     expect(nullResult, isNull);
   });
 
+  test('Rust-owned with_foreign callback can be lifted and called from Dart', () {
+    final persister = InMemoryEventPersister().asPersister();
+
+    persister.save('receiver-created');
+    persister.save('receiver-saved');
+
+    expect(persister.load(), equals(['receiver-created', 'receiver-saved']));
+    persister.close();
+  });
+
+  test('Rust-owned with_foreign callback can be passed back into Rust', () {
+    final persister = InMemoryEventPersister().asPersister();
+
+    expect(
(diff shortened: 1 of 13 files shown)


fixtures/callbacks/test/callbacks_test.dart
```diff
@@ -173,6 +173,25 @@ void main() {
...
+
+  test('Rust-owned with_foreign callback can be passed back into Rust', () {
+    final persister = InMemoryEventPersister().asPersister();
+
+    expect(
+      saveAndLoadPersister(persister: persister, event: 'payjoin-session'),
+      equals(['payjoin-session']),
+    );
+  });
+
```
spacebear21, line 194: nit: since uniffi-dart is not a payjoin library we may want to rewrite this to be more generic, like a simple counter or something like that

spacebear21 (approved):
tACK 0a1a0bacbbb419f6498c4a0168e84e309799db29

Tested against #1533 and confirmed it passes. I have one comment that can be addressed in a follow-up: we shouldn't reference payjoin (or any other specific project internals) directly in uniffi-dart tests to keep this library more generic.

PR #136, opened 2026-05-14
Test Nix dev shell setup

chavic:
This PR adds a Nix flake for reproducible (not fully reproducible yet)  local development shells and lightweight CI validation. It provides `.#msrv`, `.#stable`, and `.#nightly` shells with Rust, Dart, `cargo-nextest`, `rust-analyzer`, `cargo-watch`, `cargo-edit`, `taplo`, and `codespell`, plus direnv support through `.envrc`. The Nix CI workflow runs `nix flake check` and verifies the availability of shell tools

The setup is intentionally scoped to this repository inspired by  copying rust-payjoin. It does not add pure Cargo derivations yet because this repo does not currently commit `Cargo.lock`, and the UniFFI fixture tests dynamically generate Dart packages and run `dart test`.

A  follow-up PR would be to commit `Cargo.lock` for this workspace, remove `/Cargo.lock` from `.gitignore`, and then tighten CI with `--locked`. Once the lockfile policy is settled, we can add Cargo-focused flake checks with `crane`, such as `cargo check`, `clippy`, and no-run test builds. Full pure-Nix fixture tests should wait until Dart/Pub dependency pinning is handled.

## Testing

- `nix fmt -- --ci`
- `nix flake check -L`
- `nix develop .#msrv -c bash -lc 'rustc --version && cargo --version && cargo nextest --version && dart --version'`
- `nix develop .#stable -c bash -lc 'rustc --version && cargo --version && cargo nextest --version && dart --version'`
- `nix develop .#nightly -c bash -lc 'rustc --version && cargo --version && cargo nextest --version && dart --version'`
- `nix develop .#stable -c cargo fmt --all -- --check`
- `nix develop .#msrv -c cargo nextest run -p hello_world --nocapture`

### DIFF
--- a/treefmt.nix
+++ b/treefmt.nix
@@ -1,0 +1,8 @@
+{ ... }:
+{
+  projectRootFile = "flake.nix";
+
+  programs = {
+    nixfmt.enable = true;
+  };
+}
--- a/treefmt.nix
+++ b/treefmt.nix
@@ -1,8 +1,29 @@
-{ ... }:
+{ pkgs, ... }:
 {
   projectRootFile = "flake.nix";
 
   programs = {
+    dart-format.enable = true;
     nixfmt.enable = true;
+    rustfmt = {
+      enable = true;
+      package = pkgs.rustToolchains.stable;
+      edition = "2021";
+    };
+  };
+
+  settings = {
+    excludes = [
+      "experiments/**"
+    ];
+    formatter = {
+      dart-format.excludes = [
+        "fixtures/**/test/*.dart"
+      ];
+      rustfmt.options = [
+        "--config-path"
+        "./rustfmt.toml"
+      ];
+    };
   };
 }
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake .
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,56 @@
+name: Nix
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+
+env:
+  NIX_CONFIG: experimental-features = nix-command flakes
+
+jobs:
+  flake-check:
+    name: Flake check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v4
+
+      - name: Install Nix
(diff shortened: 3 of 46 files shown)


spacebear21 (approved):
concept ACK

treefmt.nix
```diff
@@ -0,0 +1,8 @@
+{ ... }:
+{
+  projectRootFile = "flake.nix";
+
+  programs = {
+    nixfmt.enable = true;
+  };
+}
```
benalleng: Here is a more complete formatter. We should also add the same or similar rust formatting rules from here https://github.com/payjoin/rust-payjoin/blob/master/rustfmt.toml

```suggestion
{ pkgs, ... }:
{
  projectRootFile = "flake.nix";

  programs = {
    dart-format.enable = true;
    nixfmt.enable = true;
    rustfmt = {
      enable = true;
      package = pkgs.rustToolchains.nightly;
      edition = "2021";
    };
  };
  settings = {
    formatter = {
      dart-format.options = [
        "--language-version"
        "latest"
      ]; # https://github.com/dart-lang/sdk/issues/60163#issuecomment-2668274823
      rustfmt.options = [
        "--config-path"
        "./rustfmt.toml"
      ];
    };
  };
}

```

benalleng (changes requested):
The scaffolding looks good but we do need to flesh out some the commands, the formatter specifically is mostly a no-op at this point. I think we should add a formatter for the rust and dart code but not any of the python and kotlin as most of that is just auto-generated stuff.

treefmt.nix
chavic: @benalleng  The latest commit adds some formatting rules

treefmt.nix
```diff
@@ -1,8 +1,29 @@
-{ ... }:
+{ pkgs, ... }:
 {
   projectRootFile = "flake.nix";
 
   programs = {
+    dart-format.enable = true;
```
chavic, line 6: @benalleng followed the link. Here's the result

benalleng (approved):
ACK d2af4854e2e106b3912b4907eb54d9185510436c I didn't really think about how adding this fmt command directly into this PR would really bloat the diff. Sorry for that. Looking good though 👍

PR #137, opened 2026-05-17
Complete callback interface followups

chavic:
This PR completes the callback interface followup work on top of the Rust-backed callback trait proxy support. It preserves expected callback errors separately from unexpected callback failures for both sync and async callbacks, so generated Dart callbacks now report declared UniFFI errors as `CALL_ERROR` and unexpected exceptions as `CALL_UNEXPECTED_ERROR`.

It expands active coverage across the callback fixtures. The callbacks fixture now exercises expected and unexpected callback errors, compound callback return/argument types, Rust-owned `with_foreign` callback lifting and passback, and the revived `RustStringifier` stored-callback path. The async fixture now covers expected and unexpected async callback errors and includes a second async callback interface to verify shared async helper definitions.

It also brings the proc-macro callback coverage closer to UniFFI 0.31 behavior. Proc-macro callback return types now use `Arc<dyn ...>`, the proc-macro fixture calls every callback method through Rust, and the tests cover callback values, bytes, object arguments, callback-returning-callbacks, expected errors, trait object interop, and `with_foreign` trait roundtrips.

Finally, this makes `proc-macro-no-implicit-prelude` an active workspace fixture. Its test harness now uses the current fixture runner, its callback traits use `with_foreign`, and its Dart tests are updated to the current generated API so callback interfaces are exercised without relying on Rust’s implicit prelude.

Verified with `cargo fmt --all -- --check`, Dart formatting checks, `git diff --check`, and the full `cargo test --workspace -- --nocapture` runtime suite.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ members = [
     "fixtures/trait-interfaces",
     "fixtures/dart_async",
     "fixtures/proc-macro",
+    "fixtures/proc-macro-no-implicit-prelude",
     #"fixtures/*",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,6 +77,7 @@ members = [
     "fixtures/trait-methods",
     "fixtures/trait-interfaces",
     "fixtures/dart_async",
+    "fixtures/proc-macro",
     #"fixtures/*",
 ]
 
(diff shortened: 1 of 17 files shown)


spacebear21 (approved):
utACK, test coverage going 📈

PR #124, opened 2026-04-18
fix: type helpers for aliases, compounds & callback args

Sociosarbis:
This PR fixes three issues I encountered while trying to use this project to convert https://github.com/gorules/zen/tree/master/bindings/uniffi.
1. Missing type helper for custom type aliases: When the actual underlying type of a custom type alias was not explicitly declared in the code, the type helper for that underlying type was not generated.
Example: 
```rust
pub struct JsonBuffer(pub Vec<u8>);
uniffi::custom_newtype!(JsonBuffer, Vec<u8>);
```
```dart
// actual: not generated
class FfiConverterUint8List
```
2. Missing type helper for compound inner types: Type helpers were not generated for the inner types of compound types.
Example: 
```rust
#[derive(uniffi::Record)]
pub struct ZenEngineResponse {
    pub performance: String,
    pub result: JsonBuffer,
    pub trace: Option<HashMap<String, ZenEngineTrace>>,
}
```
```dart
// actual: not generated
class FfiConverterMapStringToZenEngineTrace
```
3. Incorrect type for CallbackInterface arguments: When a method's argument was a CallbackInterface, the actual type was not used; instead, dynamic was used.
Example: 
```rust
#[uniffi::constructor]
    pub fn new(
        loader: Option<Box<dyn ZenDecisionLoaderCallback>>,
        custom_node: Option<Box<dyn ZenCustomNodeCallback>>,
    ) -> Self { ...
```
```dart
// actual
ZenEngine({required dynamic? loader,required dynamic? customNode,})
// expected
ZenEngine({required ZenDecisionLoaderCallback? loader,required ZenCustomNodeCallback? customNode,})
```
The actual effect of the changes can be verified by following the steps below. After execution, you can check the files in the build/generated/dart folder.
```bash
git clone https://github.com/Sociosarbis/zen.git
cd zen/bindings/uniffi
make build
make generate-dart
```

PR #138, opened 2026-05-17
Register recursive type helpers

chavic:
This supersedes #124 by moving helper discovery into `TypeHelpersRenderer` instead of relying on whichever render path happens to touch a type first. The renderer now walks records, enums, objects, functions, and callback interfaces through UniFFI's type iteration before taking the helper snapshot, so custom builtins and nested compound helpers are present when Dart converters are emitted.

The root cause was that helper emission was order-dependent. A custom type such as `JsonBuffer` could alias `bytes` and emit `typedef FfiConverterJsonBuffer = FfiConverterUint8List` without also registering the bytes converter, while an optional map field could emit `FfiConverterOptionalMapStringToZenEngineTrace` without emitting `FfiConverterMapStringToZenEngineTrace`.

The regression coverage reactivates `fixtures/custom_types` with a `JsonBuffer` custom bytes alias and a `ZenEngineResponse` record containing `record<string, ZenEngineTrace>?`. The Dart test exercises both lifting and lowering through exported functions, so it covers the missing custom builtin converter and the nested map converter from #124.

This branch is stacked on #137 because the callback coverage follow-up remains its base. Until #137 lands, this draft PR includes those earlier callback commits as ancestry; the new implementation commit here is `58dd9f2 Register recursive type helpers`.

Validation reproduced the pre-fix fixture failure with missing `FfiConverterUint8List` and `FfiConverterMapStringToZenEngineTrace`, then passed `cargo check -p uniffi-dart`, `cargo test -p custom_types -- --nocapture`, `cargo test -p callbacks -p proc_macro_uniffi -p proc-macro-no-implicit-prelude -- --nocapture`, and `cargo check -p custom_types`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ members = [
     "fixtures/bytes_types",
     "fixtures/callbacks",
     "fixtures/benchmarks",
+    "fixtures/custom_types",
     "fixtures/docstring-proc-macro",
     "fixtures/duration_type_test",
     "fixtures/type-limits",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ members = [
     "fixtures/trait-interfaces",
     "fixtures/dart_async",
     "fixtures/proc-macro",
+    "fixtures/proc-macro-no-implicit-prelude",
     #"fixtures/*",
 ]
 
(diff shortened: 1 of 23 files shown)


spacebear21 (approved):
utACK 58dd9f271db42348fa811637b7c9ce582a40f5ce, i guess this needs a rebase now that #137 is merged?

spacebear21 (approved):
re-ACK 6c5916b5f75adac650e8bb13693a0ce4e7e59b37

Issue #142, opened 2026-06-02
Preserve UniFFI callable argument defaults


reez:
uniffi-dart currently ignores UniFFI default values on callable arguments, so generated Dart constructors/functions require callers to pass arguments that other bindings can omit. 

Example: This shows up in bdk-dart with `ElectrumClient(validateDomain:)`: upstream bdk-ffi defaults `validate_domain = true`, but Dart requires validateDomain explicitly instead of generating `bool validateDomain = true`

https://github.com/bitcoindevkit/bdk-dart/blob/17571d0d7b4e8468464429381c7ab41fca248801/lib/bdk.dart#L21499-L21504

That forces call sites like the integration helper to pass `validateDomain: true` explicitly instead of letting Dart omit it and use the UniFFI default.

https://github.com/bitcoindevkit/bdk-dart/blob/17571d0d7b4e8468464429381c7ab41fca248801/test/integration/integration_helpers.dart#L55-L60

chavic:
resolved by #143

PR #143, opened 2026-06-02
feat: preserve callable defaults

reez:
My attempt to address #142 any/all feedback welcome

## Summary

- render UniFFI default values on generated Dart function, constructor, and method parameters
- share default rendering between record fields and callable arguments
- add coverage for omitting defaulted callable arguments


chavic (approved):
ACK

PR #144, opened 2026-06-12
Constrain toml dep to be consistent with other uniffi crates

benalleng:
I was trying to pin the transient deps within rust-payjoin to their minimal versions but I got conflicts between this and https://github.com/chavic/uniffi-bindgen-cs/pull/1 where the toml deps were all inconsistent. This attempts to lock them all to the same contraints.


spacebear21 (approved):
ACK

PR #145, opened 2026-06-18
Fixup for name mangling in all-caps acronym identifiers

jhugman:
Found in interface types, but also fixed in callback interfaces, records and custom types.

Includes a regression test in the custom types fixture.

This is not a breaking change: users who were already using the Rust naming convention e.g. `Url`, `HttpMetadata` were not seeing this bug. Users who were using all caps, e.g. `URL`, `HTTPMetadata` were getting compile errors in generated Dart.

This fixes that.


chavic (approved):
ACK

Issue #139, opened 2026-05-24
Investigate flaky dart_async timing failures


chavic:
Consolidates #105 and #123.

The `dart_async` tests can fail on clean runs due to timing-sensitive behavior in the async test suite. Before changing thresholds or rewriting assertions, we should identify the actual root cause and determine whether the failures come from test design, generated async bindings, runtime scheduling, CI load, or another layer.

Investigation scope:
- Identify which `dart_async` tests fail and under what conditions.
- Compare local, stable CI, and nightly behavior if they differ.
- Determine whether failures are caused by narrow wall-clock assertions or by a real async runtime/binding bug.
- Decide whether the fix belongs in the tests, generated Dart async code, Rust fixture behavior, CI configuration, or documentation.

The goal is to choose a fix after the root cause is understood, rather than simply relaxing timing thresholds.

PR #149, opened 2026-08-07
Support uniffi 0.32

DenisovAV:
Ports the generator from **uniffi 0.31.2 → 0.32**. The full fixture suite passes under 0.32 (`cargo test --workspace`).

### MSRV raise 1.85 → 1.88

`askama 0.16` — a hard dependency of `uniffi_bindgen` 0.32 with no 0.32-compatible older release — declares `rust-version = 1.88`, so 1.88 is the real floor. Declaring 1.85 does not help: the resolver still has to add `askama v0.16.1 (requires Rust 1.88)`, and the build then fails on 1.85.

An earlier revision of this PR raised it to 1.91. That was not a requirement but a result of unconstrained resolution: `cargo_metadata` pulled `cargo-platform 0.3.3` (rustc ≥ 1.91). With the MSRV-aware resolver from #174 (`[resolver] incompatible-rust-versions = "fallback"`), cargo falls back to `cargo-platform 0.3.2`. Verified locally on 1.88 with CI's commands: `cargo test --all --all-features` 42 passed / 0 failed, `cargo clippy --all-features --all-targets -- -D warnings` clean (after nine `uninlined_format_args` cleanups), `cargo fmt --check` clean.

The change is wired through `Cargo.toml` (`rust-version`, declared on the `uniffi-dart` crate; every member of the workspace builds through uniffi 0.32 and so needs 1.88 in practice), the CI matrix, `flake.nix` and the README.

(`toml` stays pinned to `1`: `uniffi_bindgen` 0.32.0 declares `toml = ">=0.9, <2"` and resolves to 1.x, so `new_config`/`get_toml`'s `&toml::Value` needs this crate on the same major — a looser pin splits us onto 0.9 while uniffi stays on 1.x → E0053, and there is no committed `Cargo.lock` to stabilize resolution.)

Rebased onto `main` (picks up #152, #160 and #174); the async-timing Test Suite failure was the `sleep` lower-bound flake #160 fixes, not the 0.32 path.

### What changed

- **Workspace deps** bumped to uniffi 0.32.
- **`Literal` imports** repointed from the now-private `uniffi_bindgen::pipeline::general::nodes` to the public `uniffi_bindgen::interface::Literal`.
- **`ObjectImpl`** became `Struct` + `Trait(TraitKind)` in 0.32 (the `CallbackTrait` variant was removed). Callback-interface converter naming now mirrors uniffi's own `ObjectImpl::has_callback_interface()` = `Trait(Both | ForeignOnly)`, which **preserves 0.31 behaviour**:
  - `with_foreign` traits (old `CallbackTrait`, now `TraitKind::Both`) → `FfiConverterCallbackInterface*`
  - plain trait objects (now `TraitKind::RustOnly`) → own class
- **`render_literal`** migrated off the private pipeline node types onto `interface` types. Adds the missing `Int64` numeric case (an `i64` default literal would otherwise hit the `not a number` panic) and handles the new `LiteralMetadata::EmptySet` variant — it returns no-default (matching `default_for_interface_type`) since the generator does not render `Set<T>` types yet.
- **Fixture pins**: the remaining workspace-member fixtures (`benchmarks`, `docstring-proc-macro`, `metadata`) still hardcoded `uniffi = "0.31"`; aligned them to `workspace = true`. This fixes a 0.31-writer / 0.32-reader metadata skew that broke `docstring-proc-macro` extraction (`Unexpected metadata type code`).

### New `map_type` fixture

Adds coverage for `HashMap` codegen — `String→int`, `String→record`, nested maps, optional values, and the empty-map boundary. Map is listed under **"Identified Blockers"** in the README but already generates correctly; this fixture demonstrates it and guards the Map converter's offset arithmetic against variable-length value converters.

### Notes

- Three *non-member* fixtures (`docstring`, `keywords`, `time-types`) still pin `uniffi = "0.31"`; they are not in `[workspace] members`, so they are never compiled/tested and are left untouched.
- The root crate's `+v0.31.2` build-metadata version suffix is left for a maintainer release bump.

---

### Borrowed `&[u8]` arguments → `ForeignBytes` (added)

uniffi 0.32 lowers a borrowed `&[u8]` / `[ByRef] bytes` **argument** through `FfiType::ForeignBytes` (zero-copy, foreign→Rust only) instead of the owned `RustBuffer` path — gated by `Argument::is_borrowed_bytes()`. The generator already emitted the `@Native` signature as `ForeignBytes`, but the call site still lowered the `Uint8List` with `FfiConverterUint8List.lower` (a `RustBuffer`), so **any** function taking `&[u8]` failed to compile:

```
Error: The argument type 'RustBuffer' can't be assigned to the parameter type 'ForeignBytes'.
```

The existing fixtures only use owned `Vec<u8>`, so nothing caught it, but it broke every downstream consumer with a `&[u8]` argument (e.g. payjoin-ffi's `process_response(&self, body: &[u8])`, 9 call sites — verified fixed against payjoin-ffi on 0.32: `24 tests passed`).

Fix: route borrowed-bytes args through a new `lowerForeignBytes` helper, mirroring the `FfiConverterByRefBytes` in uniffi's Kotlin/Python backends. Empty slices pass `(null, 0)` (Kotlin/Python do the same — a zero-size `calloc` is platform-variable), and there's an `Int32` length guard. Adds `take_bytes_by_ref(&[u8])` to the `bytes_types` fixture (non-empty + empty round-trips).

**Per-call leak — now fixed (addresses your review):** `lowerForeignBytes` copies into native memory (Dart's GC-managed `Uint8List` has no stable address, so unlike Kotlin's direct ByteBuffer it can't be zero-copy) and that copy was never freed — the struct + buffer leaked on every `&[u8]` call. It's now allocated from an `Arena` the call site scopes around the FFI call and released afterwards: sync calls wrap the call in `using((Arena arena) { … })` (frees on scope exit, including if the call throws); async calls hold the borrow until the Rust future settles, so the arena is released via `whenComplete`. Calls with no borrowed-bytes argument are unchanged. A functional round-trip passes despite a leak, so `borrowed_bytes_allocation_test.dart` checks the allocations directly, following the argument-rollback fixture's approach: it runs the cases in a child `dart test` against a copy of the generated bindings whose `package:ffi` import hides `calloc` and `using`, routing every native allocation through a `TrackingAllocator`. (`using` has to be replaced too: its default allocator is `package:ffi`'s own `calloc`, so hiding `calloc` alone would not reach the per-call arena.) The cases cover the free function, empty slice, void return, method, constructor, the Rust error path, and a failure while lowering a *later* argument — a new `sum_borrowed_bytes_scaled(data, scale)` fixture whose `u32` lowering throws on a negative value after `data` is already copied. Each asserts the data copy went through the tracker and nothing from the call is still live. No generator change for the test. Checked against three regressions: dropping the release from the statement-form call site fails 7 of 8 cases, from the constructor expression form exactly the constructor case, and allocating the copy from `calloc` instead of the arena (the original leak) all 8.

**Support boundaries.** Two qualifications about where borrowed `&[u8]` is reachable
today:

- **Async is defensive code.** As checked in review, uniffi 0.32.0 rejects `async` exports taking `&[u8]`,
  so the `whenComplete(_uniffiArena.releaseAll)` branch cannot be reached through a
  supported signature at present. It is emitted so the arena is still released if/when
  upstream allows async borrowed bytes, not because a current fixture exercises it.
- **Traits: Rust-only yes, foreign-implementable no.** Borrowed bytes on a Rust-only
  trait method (`TraitKind::RustOnly`) work and release the arena. On a
  foreign-implementable trait (`with_foreign`, `TraitKind::Both`/`ForeignOnly`) the
  upstream macro expansion fails before codegen is reached, so that combination is
  out of scope here rather than handled.

### Downstream CI

- **bdk-dart**: bdk-dart now pins bdk-ffi by tag `v3.1.0`, so the bdk-ffi clone moves to that commit — with mismatched versions cargo drops the `[patch]` with only a "not used in the crate graph" warning. The committed lockfile is kept (bdk-dart's script uses `--locked`); the refresh step from #174 now runs inside `bdk-dart/native`, so rustup uses the pinned toolchain rather than `stable`, and also moves `toml` to 1.x to avoid the E0053 split.
- **Toolchains**: both jobs pin 1.88.0. The downstream workspaces don't read this repo's `.cargo/config.toml`, so `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=fallback` is set at workflow level. rust-payjoin declares `rust-version` 1.85; bdk-dart and bdk-ffi declare none, in which case cargo uses the active rustc.
- Validated before pushing by running both jobs' steps locally against rust-payjoin `master` (b7c1567) and bdk-dart `main` (f8be4f4), through `dart test` (macOS host target in place of Linux): rust-payjoin 24 passed; bdk-dart 25 passed, 5 skipped.


DenisovAV:
Pushed two commits addressing the 0.32-upgrade CI failures:

- **`toml = "1"`** — 0.32's bindgen trait signatures use toml 1.x's `Value`; the old `toml = ">=0.8, <=0.9"` resolved a different `toml` crate, causing E0053 ("incompatible type for trait") on `new_config`/`get_toml` in `src/gen/mod.rs`. This was breaking the main Test Suite compile and the downstream builds.
- **MSRV 1.85 → 1.91** (`Cargo.toml` rust-version, `tests.yml` matrix, `flake.nix` msrv shell) — the 0.32 dependency tree pulls `cargo-platform 0.3.3` (via `cargo_metadata`), which requires rustc 1.91, plus `askama 0.16` (rustc 1.88). 1.91 is the real floor; verified locally with `cargo test --no-run --all`.

Verified on a fork run with Actions enabled: **Nix (all dev shells incl. msrv), Lints (1.91 / stable / nightly), and Test Suite (nightly) pass; compilation is clean on 1.91.**

Two failures remain that look like they need a maintainer call rather than a bindgen change:

1. **`Test Suite` flake** — `dart_async` (`futures_test.dart: sleep`) fails intermittently. Across reruns it jumps between the `stable` and `1.91` legs while `nightly` stays green, and the suite takes ~200s, so it reads as a timing flake rather than a version issue.
2. **`Test Downstream`** — the payjoin-ffi / bdk-ffi fixtures fail with `Unexpected metadata type code: 116`; the pinned downstream refs use an older uniffi whose metadata a 0.32 bindgen cannot read. These likely need their pins bumped to a 0.32-compatible version as part of the upgrade.

Happy to help with either if that would be useful.

chavic:
@DenisovAV thanks for working on this, i think the downstream tests should be bumped in this PR, but the flakey test needs it's own PR, i think there's an issue for it

DenisovAV:
Thanks @chavic. Addressed both of your points, plus a codegen fix the downstream turned up:

**Downstream (`payjoin` / `bdk`)** — I probed bumping them to uniffi 0.32. `payjoin-ffi` compiles and its bindings generate + run cleanly on 0.32 (**24 tests passed**) once its toolchain pin is raised (0.32 needs rustc ≥ 1.91) and its lock is refreshed for toml 1.x. But getting there surfaced a real generator bug:

- **Borrowed `&[u8]` args were lowered as `RustBuffer`, not `ForeignBytes`.** 0.32 selects `FfiType::ForeignBytes` for `is_borrowed_bytes()` args, and the `@Native` signature was already emitted as `ForeignBytes`, but the call site still used `FfiConverterUint8List.lower` → `RustBuffer` — so every function with a `&[u8]` arg failed to compile (`RustBuffer can't be assigned to ForeignBytes`). The existing fixtures only use owned `Vec<u8>`, so nothing caught it; payjoin's `process_response(&self, body: &[u8])` (9 sites) did. Fixed by routing borrowed-bytes through a `lowerForeignBytes` helper (mirrors the Kotlin/Python `FfiConverterByRefBytes`), with the empty-slice `(null, 0)` special-case and an Int32 length guard. Added `take_bytes_by_ref(&[u8])` to the `bytes_types` fixture. Details + a **known follow-up** (a per-call native leak it shares with the existing `toRustBuffer`, which needs a call-site free rather than a helper change) are written up in the PR description.

  `bdk-dart` I could not get onto 0.32 from here: it pulls `bdk-ffi` as a pinned git rev that itself pins uniffi 0.31, so it's not reachable by patching `bdk-dart`'s own manifests — that one needs its upstream refs bumped. So I'd suggest wiring the downstream job to whichever `payjoin`/`bdk` 0.32-ready refs you prefer rather than their default branches.

**Flaky `dart_async::sleep`** — agreed it's separate; opened #157 against `main`. It investigates the root cause per #139 (the binding is correct — the suite runs in ~9s locally with everything passing; on a loaded CI runner it takes ~200s and the narrow wall-clock *upper* bounds trip) and switches the delay tests to lower-bound-only, matching uniffi-rs's own `test_futures.py`. `concurrent_future` keeps its concurrency check as a relative comparison. Once #157 lands and this branch rebases, the one remaining red here (that same `sleep` flake) clears.

CI on this branch is otherwise green (Nix, Lints on 1.91/stable/nightly, Test Suite on nightly; the fixture suite incl. the new `bytes_types` coverage passes).

DenisovAV:
Hi @chavic — gentle ping on this when you get a chance 🙂 Both your review points are addressed above: the downstream bump is probed (payjoin generates + runs on 0.32, 24 tests green — which also surfaced a real `ForeignBytes` codegen bug I fixed here), and the flaky test is now its own PR, #157, per your suggestion.

Since you noted #149 lands first with #152 rebased on top, this one gates the chain — happy to rebase #152 the moment it merges. No rush, just flagging it's ready for another look. Thanks!

DenisovAV:
Hi @chavic — thanks for merging #157! An update on the downstream tests here, and a question for you.

I dug into why both jobs fail against this PR's uniffi 0.32, and it's a genuine chicken-and-egg rather than something I can cleanly fix here:

- **`rust-payjoin`** — its Dart wrapper (`payjoin-ffi/dart/native`) is a *separate* workspace that pulls `payjoin-ffi` by a pinned `rev` on `uniffi = 0.31.2`, with its own `1.85.1` toolchain pin.
- **`bdk-dart`** — re-exports upstream `bdk-ffi` (pinned rev), which exact-pins `uniffi = "=0.31.2"`.

Neither can be generated against 0.32 until its upstream migrates, and neither has a 0.32 branch yet. Downstreams can't move to 0.32 until there's a 0.32 uniffi-dart release — which this PR is what enables. Forcing them green would mean patching pinned upstream crates in CI (testing a Frankenstein, not the real consumer), which I don't think is worth doing.

Worth noting: a real 0.32-native consumer already exists — the **qdrant Edge Dart SDK** (a vector-database SDK; `qdrant/qdrant#10166`), built with exactly this stack: uniffi-dart 0.32 + the library-mode CLI (#152) + the enum-collision fix (#151). So 0.32 codegen is already proven end-to-end against a real, non-trivial consumer.

How would you like to handle the downstream jobs for the transition? A few options:

1. **Skip payjoin/bdk** (with a re-enable note) until their upstreams ship 0.32, then open the migration PRs and turn them back on.
2. **Mark them non-required** in branch protection instead — they stay honestly red but don't block. Your call, since it's repo policy.
3. **Add qdrant Edge as a 0.32-native downstream job** to replace the stuck coverage — happy to wire it up (it's a heavier build).

I'd lean toward (1) now with (3) as a follow-up, but it's your CI — what do you prefer?

src/gen/types.rs
```diff
@@ -424,6 +424,37 @@ pub fn runtime_scaffolding(ci: &ComponentInterface) -> dart::Tokens {
                 return RustBuffer.fromBytes(bytes.ref);
             }
 
+            // Lowers a `Uint8List` into a `ForeignBytes` for a borrowed `&[u8]`
+            // (`[ByRef] bytes`) argument. The Rust side only borrows the buffer
+            // for the duration of the call, so this is valid in argument
+            // position only (never lifted or read back). Dart's GC-managed
+            // `Uint8List` has no stable native address, so the bytes are copied
+            // into native memory, mirroring `toRustBuffer` above.
+            ForeignBytes lowerForeignBytes(Uint8List data) {
```
chavic: This function leaks native memory on each call. `calloc<ForeignBytes>()` and `calloc<Uint8>(length)` are never freed: Rust only borrows the buffer for the call, and the generated call site does not free it after the return. Each `&[u8]` argument call leaks its payload. payjoin-ffi has nine call sites of `process_response(&self, body: &[u8])`, so a polling loop grows without limit.

The Kotlin ByRef converter frees in a `finally` block after the call. The generated Dart call site needs the same shape. The `(null, 0)` case for an empty slice and the Int32 guard are good, but the struct allocation leaks in that path also.

Note: `toRustBuffer` above has the same latent leak. It is older than this PR and out of scope here.

chavic (commented):
Thank you. The 0.32 port is close, and the `&[u8]` / `ForeignBytes` catch is a real find that the fixtures could not show.

Two items before merge, and one note:

1. `lowerForeignBytes` leaks on each call. See the inline comment.
2. The MSRV raise from 1.85 to 1.91 is correct and necessary (askama 0.16 needs 1.88, cargo-platform needs 1.91), and it is wired through Cargo.toml, CI, and flake.nix. But it changes a compatibility promise, so it needs an explicit maintainer sign-off in the PR text, not a silent ride-along.

The Test Suite failure is the `sleep` lower-bound assertion. I measured `sleep(200)` on this branch and on `main`: both give 202 to 204 ms, so the 0.32 async path is not the cause. The test measured time with the wall clock, which NTP can move on a CI runner. #160 corrects the test. Please rebase after it merges, and the Test Suite goes green here without a change to this PR.

The downstream jobs are a repo policy question. I leave that to your discussion with @chavic above.

DenisovAV:
Thanks @chavic — both review points addressed, plus the rebase you suggested.

**1. `lowerForeignBytes` leak** (da65f67): the native copy is now allocated from an `Arena` the call site scopes around the FFI call and freed after it — `using((Arena arena) { … })` for sync (frees on scope exit, including if the call throws) and `whenComplete(arena.releaseAll)` for async (the Rust future holds the borrow until it settles). Calls with no `&[u8]` argument are unchanged. Since a functional round-trip passes even while leaking, I added proc-macro fixture exports + Dart tests covering the free-function, method, fallible (error-path) and empty-slice call sites and a 50k-call loop.

**2. MSRV 1.85 → 1.91**: now called out explicitly in the PR description (its own section) for your sign-off — cargo-platform 0.3.3 needs 1.91, askama 0.16 needs 1.88; wired through `Cargo.toml` / CI / `flake.nix`.

**3. Rebased onto `main`** for #160 — the async Test Suite `sleep` lower-bound flake is gone, so this should be green here now (the payjoin/bdk downstream jobs stay the chicken-and-egg from the thread above). One thing the rebase surfaced: `uniffi_bindgen` 0.32.0 declares `toml = ">=0.9, <2"` and resolves to toml **1.x**, so I pin `toml = "1"` to stay on its major — a looser pin let the resolver put this crate on 0.9 while uniffi stayed on 1.x (E0053 in `new_config`), which CI's fresh resolve caught since the crate has no committed `Cargo.lock`.

DenisovAV:
Update: I went ahead and greened both downstream jobs rather than leaving them as the chicken-and-egg above — **all 12 checks pass now**, including Test rust-payjoin (24 tests) and Test bdk-dart (22 pass / 5 skipped).

Both payjoin-ffi and bdk-ffi compile and run their Dart suites on 0.32 with only a version bump (no code migration), so `test-downstream.yml` now bumps them to 0.32 in-CI to exercise the real 0.32 path:
- payjoin: bump payjoin-ffi's `uniffi` to 0.32.
- bdk: clone bdk-ffi at its pinned rev, bump it, and local-`[patch]` bdk-dart onto it (+ bump bdk-dart's own `uniffi`, refresh its lock).
- both: pin the native-assets `rust-toolchain.toml` to an exact `1.91.0` (0.32's MSRV — `native_toolchain_rust` rejects `stable`) with just the Linux host target.

It's self-contained in the workflow (no forks) and commented to revert once payjoin/bdk ship 0.32 upstream. If you'd rather skip them or point the job at a different 0.32-native downstream, easy to reshape — but this keeps them honestly green against the real consumer APIs.

That leaves just the MSRV 1.85 → 1.91 sign-off (its own section in the description) — over to you on that + merge. 🙏

DenisovAV:
@chavic — status check, since everything on my side is now closed out:

- **All 12 checks green** on `50afc12`, including both downstream suites (`bdk-dart` and `rust-payjoin`) built against a forced uniffi 0.32.
- **Both review points from your Aug 31 pass are addressed**: the `lowerForeignBytes` arena leak (`da65f67`) and the collision-safe arena name plus trait-method guard (`50afc12`).
- Branch is rebased and sits **exactly on top of current `main`** — 0 commits behind, clean merge.

So rather than pinging again, one concrete question so this isn't stuck on me guessing what you're waiting for:

**Is there anything else you'd like changed before merge, or is this parked on the MSRV decision?** The PR moves `rust-version` from `1.85` to `1.91`, which 0.32 requires — that's the one change here with consequences beyond this repo, and I understand if you'd rather land it deliberately.

If MSRV is the sticking point, I'm happy to split it into its own PR so the codegen fixes can land independently. Just say which way you'd prefer and I'll do the work.

chavic:
Hi @DenisovAV was just thinking about this, I'll give it a final pass before day ends

chavic:
@DenisovAV sftlr

I don't think splitting MSRV would give it any independence materially speaking 🤔, I accept Rust 1.91 for this 0.32, I'd update the ReadMe


I think your solution to the chicken and egg issue is good enough, though the temporally downstream migrations need to be at least labeled... Probably better to have an issue for them just as a reminder 

I did find one tracking downstream main `BeyondTranslate`

DanGould:
An MSRV change for payjoin-ffi is fine, but I think this changes MSRV for the whole workspace, which certainly merits conversation outside of this PR.

If the change can be isolated I don't see a problem with including MSRV change for this one crate in this pr.

Cargo.toml
```diff
@@ -82,12 +88,13 @@ members = [
...
+    "fixtures/map_type",
     #"fixtures/*",
 ]
 
 [workspace.dependencies]
-uniffi = { version = "0.31.2" }
-uniffi_bindgen = { version = "0.31.2" }
-uniffi_build = { version = "0.31.2" }
-uniffi_testing = { version = "0.31.2" }
+uniffi = { version = "0.32" }
```
chavic, line 96: Following up on BeyondTranslate: it tracks `main` in all three generator dependency sections, but pins UniFFI to `=0.31.2` and has no committed lockfile. A fresh resolution after this merge gives it a 0.32 generator with 0.31.2 metadata. I reproduced that combination failing with `Unexpected metadata type code 97`.

Just flagging the downstream impact here. We don't need to hold this PR for a change in their repo. After we merge, a follow-up PR pinning their generator to a compatible revision would be kind; then they can move to 0.32 separately.

.github/workflows/test-downstream.yml
```diff
@@ -71,13 +89,45 @@ jobs:
...
 
           [patch.'https://github.com/Uniffi-Dart/uniffi-dart']
           uniffi-dart = { path = "../../main" }
+
+          [patch.'https://github.com/bitcoindevkit/bdk-ffi.git']
+          bdk-ffi = { path = "../../bdk-ffi/bdk-ffi" }
           EOF
+          # Lockfile pins uniffi 0.31.2 / the old bdk-ffi rev; drop it so the bump
+          # and patches re-resolve cleanly.
+          rm -f bdk-dart/native/Cargo.lock
```
chavic, line 122: The current BDK script uses `cargo build --locked` and `cargo run --locked`, so deleting the lockfile here leaves it unable to run.

Please regenerate the lockfile after applying the patches and before calling the script. I reproduced the failure without that step; with it, the unchanged script proceeds.

chavic (commented):
@DenisovAV I've finished the final pass. Thanks for the great effort and work, the arena change addresses the leak I raised, including the error and partial-allocation paths. Thanks for fixing that.

I've left a few comments on the lockfile handling, regression coverage, and the description. The wider review also found existing ownership and callback problems, but those need their own fixes, as with the older `toRustBuffer` leak.

fixtures/proc-macro/test/proc_macro_test.dart
```diff
@@ -155,5 +155,50 @@ void main() {
...
+      // Fallible variant: success path, and an error path triggered by a NON-empty
+      // input (leading 0xFF) so the throw unwinds through `using` with a real data
+      // buffer to free (not just the empty (null,0) struct).
+      expect(sumBorrowedBytesChecked(data: Uint8List.fromList([10, 20])), 30);
+      expect(
+        () => sumBorrowedBytesChecked(data: Uint8List.fromList([0xFF, 1, 2])),
+        throwsA(isA<OsExceptionBasicException>()),
+      );
+
+      // Smoke test only: many calls confirm the emitted lower+free code compiles
```
chavic, line 189: The fix works, but this loop still passes if the cleanup is removed. I'd add an allocation-aware test so we don't lose the fix later.

The counter-based tests catch that: they pass with the arena and fail when cleanup is disabled. Please cover the error path and a failure while lowering a later argument too; a successful round trip alone doesn't test either.

src/gen/oracle.rs
```diff
@@ -545,14 +547,95 @@ impl DartCodeOracle {
...
+    /// When no argument is borrowed bytes this is exactly `return call;`, so
+    /// ordinary calls keep their previous shape with zero overhead. Otherwise an
+    /// `Arena` named `_uniffiArena` (which the lowering allocates from) wraps the
+    /// call:
+    /// - sync: `using` frees the arena on scope exit, including if the call
+    ///   throws — the Rust side only borrows for the duration of the call;
+    /// - async: the Rust future holds the borrow until it settles, so the arena
+    ///   is released via `whenComplete` after the returned future completes.
+    ///
+    /// NOTE: the async+borrowed-bytes combination is currently unreachable —
```
chavic, line 600: This qualification should be in the PR description too. UniFFI 0.32.0 rejects async `&[u8]` exports, so the `whenComplete` branch is defensive code at present.

I also checked borrowed bytes on traits. Rust-only traits work and free the arena; foreign-implementable traits fail upstream macro expansion. The description should keep those support boundaries clear.

src/gen/types.rs
chavic: Checked this again on the current head. Both allocations are freed on the synchronous paths I tested, including empty input, constructors, methods, Rust errors, and failure while lowering a later argument. Disabling arena cleanup makes the allocation tests fail.

This addresses the leak I raised here. The older owned-buffer leaks remain separate, as noted in the original comment.

chavic:
To avoid scope creep, I turned other findings into separate issues and addressed the most urgent ones. I don't think any of them block this yet, aside from the review comments.

I'd like to get some of the fixes in without needing to force the 0.32 upgrade just yet...

chavic:
Along with eeariler comments, please rebase onto main to pick up #174 and replace the 1.91 MSRV bump with 1.88. I've tested this branch with the fallback resolver and Rust 1.88: all 42 workspace tests pass, including the Dart fixtures and CLI checks, without generator changes. Strict Clippy needs nine `uninlined_format_args` cleanups.

Please also drop the BDK lockfile deletion and keep the refresh step now on main after the dependency patches. Rebasing retains both steps, so the deletion needs removing explicitly.

The downstream 1.91.0 pins deserve another look too. Those builds don't inherit our resolver config, so they'll need compatible resolution and validation before lowering them to 1.88. The UniFFI 0.32 alignment and `toml = "1"` change still need to stay.

.github/workflows/test-downstream.yml
DenisovAV: Removed the deletion. Confirmed the cause: bdk-dart's `scripts/generate_bindings.sh` runs `cargo build --locked` (line 33) and `cargo run --locked` (line 36).

Keeping the lockfile exposed three more problems, so the job also changes in two places:

- bdk-dart moved to bdk-ffi tag `v3.1.0` on Sep 9. The clone of `17c48b8` (`3.1.0-alpha.0`) no longer matched, so cargo ignored the `[patch]` — that only worked before because the lockfile was deleted. `BDK_FFI_REV` now points at the `v3.1.0` commit.
- The refresh step now runs in `bdk-dart/native`. rustup reads `rust-toolchain.toml` from the working directory, so from the workspace root it resolved under `stable` and locked `cargo-platform 0.3.3`.
- It also updates `toml@0.9`. The kept lockfile holds toml 0.9.12, which `uniffi_bindgen` 0.32 would keep reusing while uniffi-dart requires toml 1 (E0053 on `new_config`).

I ran the job's steps locally against current bdk-dart `main` through `dart test` (macOS host target): 25 passed, 5 skipped.

fixtures/proc-macro/test/proc_macro_test.dart
DenisovAV: Reworked this to follow the argument-rollback approach from #180/#183 instead of touching the generator: a `TrackingAllocator`, and `borrowed_bytes_allocation_test.dart` running the cases in a child `dart test` against a copy of the generated bindings whose `import "package:ffi/ffi.dart"` hides `calloc`.

Two differences from that fixture, both needed here:

- The rewrite also hides `using`, and the tracker re-exports a `using` bound to itself. `package:ffi`'s `using` defaults to `calloc` as resolved where `using` is declared, so hiding `calloc` alone never reaches the per-call arena.
- The instrumented copy goes into `test/instrumented/` rather than rewriting the generated files in place, because `proc_macro_test.dart` loads those files concurrently in the same run.

Cases: free function, empty slice, void return, method, constructor, the Rust error path, and a failure while lowering a later argument. For the last one, a new `sum_borrowed_bytes_scaled(data: &[u8], scale: u32)` is called with `scale: -1`, so the `u32` lowering throws after `data` is already in the arena. Each case asserts the 37-byte data copy went through the tracker and that nothing allocated during the call is still live.

Checked against three regressions in the generator: dropping the release from the statement-form call site fails 7 of the 8 cases, dropping it from the constructor expression form fails exactly the constructor case, and allocating the copy from `calloc` instead of the arena (the original leak) fails all 8.

src/gen/oracle.rs
DenisovAV: Added a "Support boundaries" section to the description with both points from your check: 0.32.0 rejects async `&[u8]` exports, so the `whenComplete` branch is defensive for now; borrowed bytes work on Rust-only traits and free the arena, and fail during upstream macro expansion on foreign-implementable traits.

Cargo.toml
DenisovAV: Agreed it shouldn't hold this. Happy to open a follow-up PR on BeyondTranslate after this merges, pinning their generator to a revision compatible with their `uniffi = "=0.31.2"` so they can move to 0.32 separately.

DenisovAV:
@chavic — updated per your Sep 11 list, rebased onto `main` (including #174):

- **MSRV 1.88.** `askama 0.16`, which `uniffi_bindgen` 0.32 requires, declares `rust-version = 1.88`; the earlier 1.91 came from `cargo-platform 0.3.3`, which the #174 resolver now avoids. On 1.88: `cargo test --all --all-features` 42 passed, strict clippy clean. README updated.
- **Nine `uninlined_format_args`** fixed, no behaviour change.
- **BDK lockfile deletion removed**, refresh step kept, with two adjustments explained in that thread.
- **Downstream pins lowered to 1.88.0, validated before pushing.** I ran both jobs' steps locally against current `main` of rust-payjoin and bdk-dart through `dart test` (macOS host target in place of Linux): payjoin 24 passed, bdk-dart 25 passed / 5 skipped. The workflow sets `CARGO_RESOLVER_INCOMPATIBLE_RUST_VERSIONS=fallback`, since the downstream workspaces don't read our `.cargo/config.toml`.
- **UniFFI 0.32 alignment and `toml = "1"`** unchanged.
- **Allocation test** now follows your `TrackingAllocator` approach, with no generator change — details in that thread.

@DanGould — on isolating it: the 1.88 floor isn't something this PR can confine, because it's uniffi 0.32's own — it comes from `askama`, which `uniffi_bindgen` 0.32 depends on. Here `rust-version` is declared only on the `uniffi-dart` crate, but every member of this workspace builds through uniffi 0.32, so in practice all of them need 1.88. On the payjoin side it reaches only what builds through uniffi: `payjoin-ffi` and the `payjoin-ffi-wrapper` Dart crate. `payjoin`, `payjoin-cli`, `payjoin-mailroom`, `payjoin-test-utils` and `fuzz` don't depend on uniffi, so their 1.85 is unaffected. The toolchain overrides in this repo's downstream job only apply inside that job.

Issue #72, opened 2025-08-07
Add External Crate Support


chavic:
**Problem:**
Support for external Rust crates and cross-crate type definitions is missing.

**Acceptance Criteria:**

- [ ] Support types from external crates
- [ ] Handle cross-crate dependencies
- [ ] Generate proper imports for external types
- [ ] Support external trait implementations
- [ ] Documentation for external crate usage

PR #150, opened 2026-08-07
Emit runtime scaffolding once into a shared module

### LINKED ISSUE #84: Handle return errors from a different namespace
See the failed run from https://github.com/payjoin/rust-payjoin/actions/runs/17413540585/job/49436359381#step:7:54 for an example.

The function returns an error type from a different FFI namespace:

```rust
    pub fn with_max_fee_rate(
        &self,
        max_effective_fee_rate_sat_per_vb: u64,
    ) -> Result<Self, bitcoin_ffi::error::FeeRateError> {
```

which results in a type error due to `UniffiRustCallStatusErrorHandler` expecting a `RustBuffer` from the current namespace:

```
00:00 +0 -2: loading test/test_payjoin_integration_test.dart [E]
  Failed to load "test/test_payjoin_integration_test.dart":
  lib/payjoin_ffi.dart:4902:13: Error: The argument type 'FeeRateExceptionErrorHandler' can't be assigned to the parameter type 'UniffiRustCallStatusErrorHandler?'.
   - 'FeeRateExceptionErrorHandler' is from 'lib/bitcoin.dart'.
   - 'UniffiRustCallStatusErrorHandler' is from 'lib/payjoin_ffi.dart'.
          )), feeRateExceptionErrorHandler);
              ^
```

1egoman:
## Problem

Issue #84 reports a type error in generated Dart code due to a Rust function returning an error type from a different UniFFI namespace. The generated Dart code then does not compile:

```
lib/payjoin_ffi.dart:4902:13: Error: The argument type 'FeeRateExceptionErrorHandler'
can't be assigned to the parameter type 'UniffiRustCallStatusErrorHandler?'.
 - 'FeeRateExceptionErrorHandler' is from 'lib/bitcoin.dart'.
 - 'UniffiRustCallStatusErrorHandler' is from 'lib/payjoin_ffi.dart'.
```

## Cause

The generator wrote the runtime scaffolding into every component file - each file declared its own `UniffiRustCallStatusErrorHandler.` Dart creates a separate type for each declaration, and two component files therefore held two different types with the same name. An error handler from one file cannot pass to `uniffiRustCallAsync` in another file.

`RustBuffer` had the same problem, and #50 corrected `RustBuffer` only. It imported each external namespace two times, and one import used a library prefix. The generator then wrote that prefix before each external `RustBuffer`. This method corrects one type. It does not correct the error handler types, or any other runtime type.

## Change

The generator now emits the runtime scaffolding one time, into a new file named `uniffi_runtime.dart`. Every component file imports this file, and every component file also re-exports it. The re-export keeps the public API complete for a consumer that imports only the component file.

The `rustbuffer_*` extern declarations move into the new file, because the methods of RustBuffer call them. `uniffi_function_definitions` therefore takes a predicate. The predicate divides the FFI functions between the runtime file and the component files. A symbol must not have a declaration in two files.

The change has two further results:

- `RustBuffer` is now a single type. The namespace prefix from #50 is no longer necessary, so this PR removes it.
- The foreign-future helpers lose the underscore at the start of their names. Generated component code refers to these helpers. Dart does not make a private name visible outside its own library.

## Tests

Please note the limit of the coverage of existing tests. The fixtures run in UDL mode, and each fixture generates one component. No fixture generates two components together. The fixtures therefore show that this change causes no regression. They do not prove the correction.

I tested the multi-component case by hand. I generated bindings for two crates, livekit-uniffi and livekit-datatrack. The second crate re-exports the first. Before this change, the generated Dart code did not compile. After this change, it compiles.

A fixture with two components is a good follow-up task - let me know and I can try to add something here. Such a fixture would protect this behavior and would also help #72.

Fixes #84.


1egoman:
I just pushed ab28306 to attempt to fix some clippy errors, but it looks like I can't be sure that lets that ci job pass until a maintainer approves the ci run for this commit.

There was also a failing test in another ci job - from what I can tell, this seems to be a known flaky one (https://github.com/Uniffi-Dart/uniffi-dart/issues/139) so I'll leave this unaddressed.

chavic:
@1egoman, thanks for working on this. I'm taking a look now

src/gen/mod.rs
```diff
@@ -82,6 +82,92 @@ impl Config {
     }
 }
 
+/// The `RustBuffer` alloc/free/reserve helpers, which `RustBuffer`'s own methods
+/// call. They live alongside `RustBuffer` in the shared runtime rather than being
+/// repeated in every component file.
+fn is_rustbuffer_fn(name: &str) -> bool {
```
chavic, line 88: **This item stops the compilation.**

`is_rustbuffer_fn` looks for the text `_rustbuffer_` in the name. Because of this, it also finds the functions of the user.

I added this function to `demotop`:

```rust
#[uniffi::export]
pub fn to_rustbuffer_bytes(n: u32) -> u32 { n + 1 }
```

Then the Dart code did not compile:

```
error - lib/demotop.dart:44:17 - The function
'uniffi_demotop_fn_func_to_rustbuffer_bytes' isn't defined
error - lib/demotop.dart:428:7 - The function
'uniffi_demotop_checksum_func_to_rustbuffer_bytes' isn't defined
```

The runtime file gets the functions of the first component only. Because of this, the extern declaration goes to no file.

A comparison of the full names prevents this:

```rust
fn is_rustbuffer_fn(ci: &ComponentInterface, name: &str) -> bool {
    [ci.ffi_rustbuffer_alloc(), ci.ffi_rustbuffer_from_bytes(),
     ci.ffi_rustbuffer_free(), ci.ffi_rustbuffer_reserve()]
        .iter().any(|f| f.name() == name)
}
```

`runtime_scaffolding` calls these same four functions, and then the two lists agree.

src/gen/types.rs
```diff
@@ -181,11 +187,96 @@ impl Renderer<(FunctionDefinition, dart::Tokens)> for TypeHelpersRenderer<'_> {
             import "dart:isolate";
             import "dart:typed_data";
             import "package:ffi/ffi.dart";
+            $(format!("import \"{RUNTIME_MODULE}.dart\";"))
+            // Re-exported so a consumer importing only this file still sees the
+            // shared runtime types that surface in the generated public API.
+            $(format!("export \"{RUNTIME_MODULE}.dart\";"))
             $(imports)
```
chavic, line 194: This is a small difference from the text of the PR. The change removes the user of the prefix in `oracle.rs`, but `$(imports)` here still writes the second import. The source is lines 155-160 above.

Output:

```dart
import "demoext.dart";
import "demoext.dart" as demoext;
```

Dart gives a warning:

```
warning - lib/demotop.dart:14:8 - Unused import: 'demoext.dart'
```

The second import and the comment above it can go out with the prefix.

src/gen/oracle.rs
```diff
@@ -114,12 +114,9 @@ impl DartCodeOracle {
     }
 
     /// Helper method to fully qualify imports of external `RustBuffer`s
-    fn rust_buffer_name_with_path(module_path: &str, ci: &ComponentInterface) -> dart::Tokens {
-        let namespace =
-            ci.namespace_for_module_path(module_path).expect("module path should exist");
-        if namespace != ci.namespace() {
-            return quote!($(namespace).RustBuffer);
-        }
+    fn rust_buffer_name_with_path(_module_path: &str, _ci: &ComponentInterface) -> dart::Tokens {
```
chavic, line 117: After this change, `rust_buffer_name_with_path` has two unused parameters and returns a constant. The comment above it describes the earlier behavior.

The function can go out. Then `quote!(RustBuffer)` can go at the three lines that call it: 111, 354 and 398.

src/gen/types.rs
```diff
@@ -181,11 +187,96 @@ impl Renderer<(FunctionDefinition, dart::Tokens)> for TypeHelpersRenderer<'_> {
             import "dart:isolate";
             import "dart:typed_data";
             import "package:ffi/ffi.dart";
+            $(format!("import \"{RUNTIME_MODULE}.dart\";"))
```
chavic, line 190: The runtime code moved out of the component file, but the import lines above stay.

For `hello_world`, the number of `dart analyze` problems increases from 3 to 8. For my two crates, the number is 14.

```
warning - lib/hello_world.dart:3:8 - Unused import: 'dart:async'
warning - lib/hello_world.dart:4:8 - Unused import: 'dart:convert'
... (6 in total, and 2 on main)
```

`dart analyze` fails when it finds a warning. Because of this, a user who does an analysis of the generated code gets a failure.

One method is to write this line at the top of each generated file:

```dart
// ignore_for_file: unused_import
```

A second method is to record the imports that the file uses. That method is more work than the result gives.

src/gen/mod.rs
```diff
@@ -241,6 +256,32 @@ impl BindingGenerator for DartBindingGenerator {
...
 
             tokens.format_file(&mut w.as_formatter(&fmt), &config)?;
+            Ok(())
+        };
+
+        // The shared runtime goes out once for the whole generation, ahead of the
+        // per-component files that import it. Any component supplies the
+        // `rustbuffer_*` symbols and asset id -- they all resolve into the same
+        // cdylib, so the choice is arbitrary.
+        if let Some(Component { ci, config, .. }) = components.first() {
```
chavic, line 266: The runtime file uses the `rustbuffer_*` symbols and the asset id of the first component. All the other components use them also.

This is correct now, for three reasons:

- Library mode is the only mode with more than one component.
- All the components come from one cdylib.
- UniFFI gives the same cdylib name to each component.

My test made 2000 allocations across two namespaces. All of them were correct.

But only a comment holds this rule. A test in the code gives more safety. Use `bail!` if two components have different asset ids. Then a change to the configuration gives an error, and does not give incorrect code.

One more small item: a crate with the namespace `uniffi_runtime` writes its component file over the runtime file.

src/gen/enums.rs
```diff
@@ -36,7 +36,7 @@ impl CodeType for EnumCodeType {
     }
 
     fn ffi_converter_name(&self) -> String {
-        format!("FfiConverter{}", &DartCodeOracle::class_name(&self.id))
+        format!("FfiConverter{}", DartCodeOracle::class_name(&self.id))
```
chavic, line 39: The changes to the format arguments are correct, and they are safe.

I tried to reproduce the clippy failure and I could not. I ran `cargo clippy -- -D warnings`, which is the command in CI. On `main`, clippy reported no problem. My rustc version is 1.94.0-nightly. The CI for `main` is also green.

Maybe your toolchain has a different configuration. Can you give the clippy version and the output? If the failure is not on `main`, then this commit can go in a different PR, or the message can change.

There is one more item, and it is not related to this commit. `cargo clippy --all-targets -- -D warnings` fails on `main` and on this branch with `items_after_test_module`. This change moves `mod tests` to a higher position in `types.rs`. Because of this, `generate_type` and `runtime_scaffolding` come after the test module. CI does not use `--all-targets` and does not find this problem. A move of the test module to the end of the file corrects it.

chavic (commented):
Thank you for this change. The method is correct.

I tested the change. I made two crates:

- `demoext` gives a record and an error.
- `demotop` gives a function that returns the record or the error.

I made the bindings in library mode. Then I did an analysis of the Dart code.

On `main`, the Dart code has four errors. These are the errors in issue #84:

```
error - lib/demotop.dart:423:5 - The argument type 'ExtExceptionErrorHandler' can't be
        assigned to the parameter type 'UniffiRustCallStatusErrorHandler?'
error - lib/demotop.dart:433:5 - ... 'RustBuffer' is defined in demoext.dart / demotop.dart
```

With this change, the Dart code has no errors.

The Dart tests are satisfactory. The tests include these operations:

- A call across two namespaces.
- An error across two namespaces.
- 2000 RustBuffer allocations across two namespaces.
- An async call across two namespaces.

I found seven items. One item stops the compilation. The other six are small, and most of them are cleanups in the same area. Six items are in the code, below.

## The tests

I agree with your note about the fixtures.

The risk of this change occurs only when one generation makes two components. Because of this, a test with two components gives more protection than the other fixtures give.

My test setup is small:

- Two crates with proc-macro exports.
- One call to `generate_dart_bindings` in library mode.
- A test that `dart analyze` finds no error.

I am happy to make this fixture and send it to you, if that helps.

src/gen/mod.rs
1egoman: Yes, very good point. In hindsight, I'm not sure this was the best way to accomplish this. I'll migrate to what you have suggested.

~Is there a list of all of these `ffi_rustbuffer_*` functions available in a constant somewhere the codebase I could pull in? IMO It would be nice to avoid the extra duplication if so.~

EDIT: Ah, I discovered [`ci.iter_rust_buffer_ffi_function_definitions()`](https://docs.rs/uniffi_bindgen/latest/uniffi_bindgen/interface/struct.ComponentInterface.html#method.iter_rust_buffer_ffi_function_definitions) - exactly what I was looking for. I will use this instead.

1egoman: Ah, I discovered [`ci.iter_rust_buffer_ffi_function_definitions()`](https://docs.rs/uniffi_bindgen/latest/uniffi_bindgen/interface/struct.ComponentInterface.html#method.iter_rust_buffer_ffi_function_definitions) - exactly what I was looking for. I will use this instead.

1egoman: Fixed in e43aed28a7c2c8b0082d2b7bdb9f10656111485e

src/gen/types.rs
1egoman: Makes sense - fixed in be051ef. The generator now writes one import for each external namespace. The comment above it has also been dropped.

src/gen/oracle.rs
1egoman: Sounds good - dropped `rust_buffer_name_with_path` and the three lines now use `quote!(RustBuffer)`. `rust_buffer_name` above it is also gone, because it was no longer being used, along with the`ComponentInterface` parameter (which only did the namespace lookup) from five of the functions, and the `renderer` parameter from two callback-interface helpers. The generated Dart did not change. Fixed in 6e54745

src/gen/types.rs
1egoman: Admittedly, my dart knowledge is not super deep, so I opted for your first proposed method as I think it will be a lot easier. The build hook of the test tools also imported `dart:io` and did not use it, so I also dropped that import. `dart analyze` on the `hello_world` output now finds no issues. Fixed in 177d77f

src/gen/mod.rs
1egoman: Good point, I missed these cases and agree a hard error makes sense here:
- `write_bindings` now sends an error if two components give different asset ids,
- An error is raised if the namespace of a component is `uniffi_runtime`, because that component file writes over the runtime file as mentioned.

Fixed in 094d3ce

src/gen/enums.rs
1egoman: I moved the test module to the end of types.rs. `cargo clippy --all-targets -- -D warnings` is now clean for me - fixed in 31951aa.

Going back and forth with a LLM, it looks like the reason you can't reproduce this failure is due to toolchain version. CI seems to use 1.97.0 which causes it to occur, and if I use `1.94.0-nightly` (or what I was using initially in development, rustc `1.94.1` / clippy `0.94.1`) it all looks fine to me locally as well. The reason it seems to not fail on `main` is because the last commit was ~2 months ago (which ran on rustc 1.96.0), and I suspect that if you re-ran the CI again now it _would_ fail.

What version of rust should these lints be expected to pass on? Should that be the version that is currently used by ci (1.97.0), or should the version running in ci be changed to be fixed at `1.96.0` /  `1.94.0-nightly` release and this extra ab28306 commit dropped?

1egoman:
@chavic Thanks for the thorough review! I believe I've addressed all of your notes and left a question on the lint related comment, would you be able to take another look?

chavic (approved):
Nice work on the fixes, thanks! Verified everything locally — just needs the rebase once #158 lands.

1egoman:
@chavic Rebased, you should be good to go!

Also rebased the other PRs that I have opened - https://github.com/Uniffi-Dart/uniffi-dart/pull/156 and https://github.com/Uniffi-Dart/uniffi-dart/pull/154, which I think should also be fully ready for review at this point.

PR #152, opened 2026-08-08
Implement the library-mode bindgen CLI (uniffi_bindgen_dart)

DenisovAV:
### What

`src/bin.rs` was a placeholder (`"This binary is a placeholder and will be implemented in the future."`). This wires it up to the **already-present** library-mode entry point, `generate_dart_bindings(.., library_mode: true)` (which calls `uniffi_bindgen::library_mode::generate_bindings` with `DartBindingGenerator`). No generator logic is added — just the CLI in front of it:

```
uniffi_bindgen_dart --library <lib.{so,dylib,dll}> --out-dir <dir> [--config <toml>] [--udl <udl>]
```

### Why

Crates that use `uniffi::setup_scaffolding!()` (proc-macro, **no UDL**) can only be consumed in **library mode** — the interface is read from the compiled library's embedded metadata. The existing `generate_scaffolding` / build.rs path requires a UDL, so there was no way to generate bindings for a pure proc-macro crate from the CLI. This closes that gap.

Verified end-to-end by generating bindings for a real ~90-type `setup_scaffolding!()` crate; the output passes `dart analyze` with no errors.

### Details

- std-only arg parsing, no new dependencies.
- `--udl` is optional: in library mode the interface comes from the library metadata, so it defaults to an unread placeholder for proc-macro crates.
- Requires the `binary` feature (as configured by the existing `[[bin]]` `required-features`).

### Note

Includes a small `chore` commit dropping a redundant `&` in a few `format!`/`println!` args — stable clippy 1.97+ added `redundant reference in format! argument`, which fails `cargo clippy -- -D warnings` on the current CI for pre-existing code. Needed for green CI; harmless to drop on rebase if it lands elsewhere first.


chavic (commented):
Thank you for this change. About time this got done; a CLI for library mode is necessary.

I compared it with `uniffi-bindgen` (uniffi-rs 0.31.2), `uniffi-bindgen-cs`, and `uniffi-bindgen-go`. The three agree with each other, and this CLI is different from all three. Two of the differences make incorrect output with no error message. The comments below give the details.

One note about the order of the merges. PR #149 moves the repository to uniffi 0.32, and the `--config` file has a new format in 0.32. Please put #149 first, and rebase this PR on it. #149 does not touch `src/bin.rs` or the config supplier, so the two changes do not conflict.

src/bin.rs
```diff
@@ -1,4 +1,77 @@
...
+        exit(2);
+    };
+
+    // In library mode the interface is read from the compiled library's embedded
+    // metadata; a UDL path is only consulted for UDL-based crates. Default it to
+    // the library path (never read for proc-macro crates) so pure
+    // `setup_scaffolding!()` crates need no `--udl`.
+    let udl = udl.unwrap_or_else(|| library.clone());
+
+    if let Err(e) = generate_dart_bindings(
```
chavic, line 65: Without `--config`, the CLI ignores the `uniffi.toml` of the crate, because `LocalConfigSupplier` (`src/gen/mod.rs`, line 381) does not read it.

The result is silent. `package_name` and `cdylib_name` fall back to defaults, so the asset id does not agree with the id in the Native Assets build hook. The Dart code compiles, and then it does not find the library at run time.

The other generators read the `uniffi.toml` of each crate, and UniFFI supplies the code for this:

```rust
Box::new(uniffi_bindgen::cargo_metadata::CrateConfigSupplier::from_cargo_metadata_command(false)?)
```

Use `from_cargo_metadata_command`. Do not use `From<Metadata>`: this repository uses cargo_metadata 0.19 and uniffi_bindgen uses 0.18, and the two versions give a confusing type error.

This supplier is in 0.31.2 and in 0.32.0, so the change is safe before and after the move to 0.32.

src/bin.rs
```diff
@@ -1,4 +1,77 @@
...
+    let (Some(library), Some(out_dir)) = (library, out_dir) else {
+        eprintln!("error: --library and --out-dir are required\n\n{USAGE}");
+        exit(2);
+    };
+
+    // In library mode the interface is read from the compiled library's embedded
+    // metadata; a UDL path is only consulted for UDL-based crates. Default it to
+    // the library path (never read for proc-macro crates) so pure
+    // `setup_scaffolding!()` crates need no `--udl`.
+    let udl = udl.unwrap_or_else(|| library.clone());
```
chavic, line 63: In library mode, UniFFI finds the UDL of each crate from the crate name. One `--udl` option gives one path for all the crates, so a library with two UDL crates cannot get a correct result.

The default value here is the path of the library. For a UDL crate with no `--udl`, the supplier reads the `.so` as UDL text and reports a UTF-8 error that does not name the missing option.

`CrateConfigSupplier` gives each crate its own UDL. Then this line and the `--udl` option can go out.

Note: UDL is not obsolete. UniFFI 0.32 keeps it, and #149 adds a new UDL fixture. Only the method to find the UDL is different.

src/bin.rs
```diff
@@ -1,4 +1,77 @@
...
+
+use camino::Utf8PathBuf;
+
+use uniffi_dart::gen::generate_dart_bindings;
+
+const USAGE: &str = "\
+uniffi_bindgen_dart — generate Dart bindings from a compiled UniFFI library
+
+USAGE:
+    uniffi_bindgen_dart --library <LIB> --out-dir <DIR> [--config <TOML>] [--udl <UDL>]
```
chavic, line 11: In `uniffi-bindgen`, `uniffi-bindgen-cs`, and `uniffi-bindgen-go`, the source path is positional and `--library` is a boolean. In uniffi-rs 0.31.2 `--library` is also deprecated, because UniFFI finds the type of the source itself.

A user who knows `uniffi-bindgen` writes the path after the options, and this parser then reports `unknown argument`. Please use the same shape. A change after the release is a break for the users.

Two options from the other tools are absent:

- `--crate <NAME>` selects one crate from a library with more than one component. Now the tool always makes all of them.
- `--no-format` stops the call to `dart format`.

Please also tell the user that the command runs `cargo metadata` in the current directory, and fails outside a Cargo workspace.

src/bin.rs
```diff
@@ -1,4 +1,77 @@
...
+
+OPTIONS:
+    --library <LIB>   Path to the compiled cdylib/dylib (.so/.dylib/.dll) whose
+                      embedded UniFFI metadata is read (library mode). This is how
+                      `setup_scaffolding!()` (proc-macro, no UDL) crates are consumed.
+    --out-dir <DIR>   Directory the generated `.dart` binding is written into.
+    --config <TOML>   Optional `uniffi.toml` config file.
```
chavic, line 18: `--config` replaces the per-crate configuration instead of adding to it, because `ConfigFileSupplier::get_toml` ignores `crate_name` and returns the same file for every crate. With more than one component they all get the same asset id, and only one of them finds its library at run time.

In UniFFI, `--config` is merged with the `uniffi.toml` of each crate, and its values have precedence. `CrateConfigSupplier` gives this behavior.

The format of the file also changes in 0.32: it becomes a global config with `[defaults]`, `[crates.<name>]`, and `[crate-roots]`. This is the reason for the rebase on #149 — then `--config` has the new format at the start.

src/bin.rs
```diff
@@ -1,4 +1,77 @@
...
-    println!("This binary is a placeholder and will be implemented in the future.");
+    let mut library: Option<Utf8PathBuf> = None;
+    let mut out_dir: Option<Utf8PathBuf> = None;
+    let mut config: Option<Utf8PathBuf> = None;
+    let mut udl: Option<Utf8PathBuf> = None;
+
+    let mut args = std::env::args().skip(1);
+    while let Some(arg) = args.next() {
+        let mut take = |name: &str| {
+            args.next().map(Utf8PathBuf::from).unwrap_or_else(|| {
```
chavic, line 33: Three small items about the parser. None of them is important.

- `take` accepts a value that starts with `--`, so `--library --out-dir x` makes the library path `--out-dir`.
- A repeated option keeps the last value and gives no message.
- There is no `--version` option.

src/gen/render/mod.rs
```diff
@@ -56,9 +56,7 @@ pub trait Renderable {
             _ => todo!("Type::{:?}", ty),
         };
 
-        if !type_helper.include_once_check(&ty.as_codetype().canonical_name(), ty) {
-            println!("{} Added", &ty.as_codetype().canonical_name());
-        }
+        type_helper.include_once_check(&ty.as_codetype().canonical_name(), ty);
```
chavic, line 59: The removal of this line is good. The `{name} Added` text made the output difficult to read.

One note: #149 and #150 both change this same line, and this PR removes it. The correction at merge is easy, but the last two PRs to land must do it.

DenisovAV:
Thanks for the thorough review @chavic — really helpful. Pushed two commits addressing it:

**Config supplier (the two silent bugs).** Replaced `LocalConfigSupplier` / `ConfigFileSupplier` with `uniffi_bindgen::cargo_metadata::CrateConfigSupplier::from_cargo_metadata_command`, and pass the `--config` override through `library_mode::generate_bindings`' `config_file_override` parameter (it was hardcoded `None`). This fixes both:

- without `--config`, the crate's `uniffi.toml` is now read, so `package_name`/`cdylib_name` no longer default and desync the asset id;
- `--config` is now *merged* per crate (values take precedence) instead of replacing every crate's config with the same file.

Verified against a crate whose `uniffi.toml` sets `package_name = "my_custom_pkg"`: the CLI in library mode with no `--config` now emits `_uniffiAssetId = "package:my_custom_pkg/uniffi:…"` (previously the default `uniffi`). As you noted, `CrateConfigSupplier` is in both 0.31.2 and 0.32, so this is safe pre- and post-rebase.

**CLI shape.** Made the source path positional and `--library` an accepted-but-optional boolean (matching `uniffi-bindgen`/`-cs`/`-go`), so writing the path after the options no longer errors. Removed `--udl` (the supplier resolves per-crate UDL now). Added `--crate <NAME>` and `--no-format`, and hardened the parser: an option value starting with `--` is rejected, a repeated option errors instead of keeping the last value, and `--version` is added. The help text notes it runs `cargo metadata` in the cwd.

**Sequencing.** Agreed on landing #149 first and rebasing this on it — the supplier/CLI changes here don't touch anything #149 changes, and the `--config` *file format* difference (the global `[defaults]`/`[crates.<name>]` shape in 0.32) is data handled by the version-appropriate `CrateConfigSupplier`, so no code change is needed here for it. Happy to rebase as soon as #149 is in. I also left the `render/mod.rs` `{name} Added` line as-is per your note (it's the same line #149/#150 touch — last one to land resolves it).

chavic (approved):
Nice work, thanks! All verified locally — just drop the clippy commit once #158 lands.

DenisovAV:
Done — dropped the clippy commit now that #158 is on `main`, rebased on top. One incidental conflict in `render/mod.rs`: #158 touched the `&` in the `println!("… Added")` debug line, which this PR removes entirely (keeping the CLI's stdout clean) — kept the removal. Green locally: `clippy --all-features --all-targets -- -D warnings` and `cargo test --no-run --all-features`. Ready for another look 🙂

chavic:
The two downstream failures are not the flake. The new `crate_name` and `try_format` parameters change the signature of `generate_dart_bindings`, and payjoin-ffi calls this function directly from its `uniffi-bindgen.rs`:

    error[E0061]: this function takes 7 arguments but 5 arguments were supplied

This will need some care to ensure downstream doesn't break; this is also probably a good chance to follow a loose standard for how third-party bindings are dealing with CLI implementation

DenisovAV:
Reshaped per your review — downstream is green now (both payjoin-ffi and bdk-dart).

- Restored the 5-arg `generate_dart_bindings` as a stable shim, so both compile unchanged.
- `crate_name`/`try_format` moved behind a `#[non_exhaustive] DartBindgenOptions` (`generate_dart_bindings_with_options`) — future knobs won't touch the signature again.
- Exposed the CLI as `uniffi_dart::main()` (in `src/cli.rs`, `binary`-gated) — the convention `uniffi::uniffi_bindgen_main()` / `uniffi_bindgen_cs::main()` use, so downstreams can forward argv instead of the raw fn. I think that's the loose standard you meant.
- Also gated the external `dart format` on `try_format` (`--no-format` was only half-honored), and added `tests/cli.rs` for the arg parser.

chavic:
Nice

PR #154, opened 2026-08-10
Lower custom types to their builtin in FFI signatures

### LINKED ISSUE #153: Custom types are emitted by their Dart alias in FFI signature positions
## Summary

The generator writes the Dart name of a custom type into FFI signature positions. A custom type becomes a Dart `typedef` - since a `typedef` is not a `NativeType`, Dart rejects the signature, and the generated file does not compile.

The defect appears in callback interfaces, because those are the positions where the generator writes a native function type by hand.

## Reproduction

This UDL is sufficient to reproduce:

```webidl
[Custom]
typedef string CustomId;

callback interface CustomIdTransformer {
    void on_event(CustomId id);
    CustomId current_id();
};

namespace example {
    void register(CustomIdTransformer observer);
};
```

## Actual generated Dart

The custom type becomes an alias of `String`:

```dart
typedef CustomId = String;
```

The generator then writes that alias into both FFI signature positions. The argument position comes from `on_event`, and the return position comes from `current_id`:

```dart
typedef UniffiCallbackInterfaceCustomIdTransformerMethod0 =
    Void Function(Uint64, CustomId, Pointer<Void>, Pointer<RustCallStatus>);

typedef UniffiCallbackInterfaceCustomIdTransformerMethod1 =
    Void Function(Uint64, Pointer<CustomId>, Pointer<RustCallStatus>);
```

Dart rejects both forms:

```
error - 'CustomId' doesn't conform to the bound 'NativeType' of the type
        parameter 'T'. Try using a type that is or is a subclass of
        'NativeType'. - type_argument_not_matching_bounds
        - The inverted type 'Pointer<String>' is also not regular-bounded, so
          the type is not well-bounded.

error - The type 'M0' given to 'fromFunction' must be a valid 'dart:ffi'
        native function type. - must_be_a_native_function_type
```

The second error stops `Pointer.fromFunction`. The generator uses that call to pass a Dart callback to Rust, so no callback interface with a custom-type argument can work.

## Expected behavior

A custom type is an alias at the Dart level only - its builtin crosses the FFI. For a custom type over `string`, the FFI type is `RustBuffer`:

```dart
typedef UniffiCallbackInterfaceCustomIdTransformerMethod0 =
    Void Function(Uint64, RustBuffer, Pointer<Void>, Pointer<RustCallStatus>);

typedef UniffiCallbackInterfaceCustomIdTransformerMethod1 =
    Void Function(Uint64, Pointer<RustBuffer>, Pointer<RustCallStatus>);
```

## Notes

The value-level converter is already correct. The generator emits this line for the same type:

```dart
typedef FfiConverterCustomId = FfiConverterString;
```

The converter therefore delegates to the builtin, and only the FFI type label is incorrect. The two paths must give the same answer.

The relevant code is in `src/gen/oracle.rs`. Two functions handle `Type::Custom` by name instead of by builtin: `native_type_label` and `native_dart_type_label`.

1egoman:
Fixes #153 

## Summary

`native_type_label` and `native_dart_type_label` handled `Type::Custom` by name. They emitted the Dart class name of the custom type, which is a `typedef`. A `typedef` is not a `NativeType`, so Dart rejected the signature.

Both functions now recurse into `builtin`. A custom type over `string` therefore lowers to `RustBuffer`, which is the type that actually crosses the FFI.

## Effect on generated code

For a custom type `DataTrackSid` over `string`, used in a callback interface:

```diff
 typedef UniffiCallbackInterfaceCustomIdTransformerMethod0 =
-    Void Function(Uint64, CustomId, Pointer<Void>, Pointer<RustCallStatus>);
+    Void Function(Uint64, RustBuffer, Pointer<Void>, Pointer<RustCallStatus>);

 typedef UniffiCallbackInterfaceCustomIdTransformerMethod1 =
-    Void Function(Uint64, Pointer<CustomId>, Pointer<RustCallStatus>);
+    Void Function(Uint64, Pointer<CustomId>, Pointer<RustCallStatus>);
```

## Tests

The second commit adds coverage for this defect to the `custom_types` fixture. `CustomIdTransformer.transform` takes a custom type as its argument and returns it, so the generator must lower the type in both FFI signature positions. The Dart test then asserts both directions of the round trip.

The test fails without the first commit:

```text
Error: Type argument 'CustomId' doesn't conform to the bound 'NativeType' of the
       type variable 'T' on 'Pointer'.
Error: The argument type 'String' can't be assigned to the parameter type 'RustBuffer'.
Error: Expected type 'NativeFunction<Void Function(Uint64, String,
       Pointer<String>, Pointer<RustCallStatus>)>' to be a valid and
       instantiated subtype of 'NativeType'.
    Pointer.fromFunction<UniffiCallbackInterfaceCustomIdTransformerMethod0>(
```

### DIFF
--- a/fixtures/custom_types/test/custom_types_test.dart
+++ b/fixtures/custom_types/test/custom_types_test.dart
@@ -3,6 +3,18 @@ import 'dart:typed_data';
 import 'package:test/test.dart';
 import '../custom_types.dart';
 
+/// Records what Rust passes in, so the test can assert on both directions of the
+/// round trip.
+class UpperCaseCustomIdTransformer extends CustomIdTransformer {
+  final List<CustomId> received = <CustomId>[];
+
+  @override
+  CustomId transform(CustomId id) {
+    received.add(id);
+    return id.toUpperCase();
+  }
+}
+
 void main() {
   test('custom alias bytes and nested map helpers round trip', () {
     final response = getZenEngineResponse();
@@ -43,5 +55,22 @@ void main() {
     expect(roundTrip.primary.status, equals(200));
     expect(roundTrip.fallback!.url, equals('https://fallback.example'));
     expect(roundTrip.fallback!.status, equals(503));
+  });
+
+  // A custom type in a callback interface signature must lower to its builtin.
+  // The alias is a Dart typedef, and a typedef is not a `NativeType`, so the
+  // generated file does not compile if the generator emits the alias here.
+  test('custom type survives a callback interface round trip', () {
+    final transformer = UpperCaseCustomIdTransformer();
+
+    final result = roundtripCustomIdThroughCallback(
+      transformer: transformer,
+      id: 'id-abc123',
+    );
+
+    // Rust passed the argument to Dart without a change.
+    expect(transformer.received, equals(<String>['id-abc123']));
+    // Dart returned a value, and Rust passed it back.
+    expect(result, equals('ID-ABC123'));
   });
 }
--- a/fixtures/custom_types/src/api.udl
+++ b/fixtures/custom_types/src/api.udl
@@ -26,8 +26,22 @@ dictionary APIResult {
     HTTPMetadata? fallback;
 };
 
+// A custom type used in callback interface signature positions. A callback
+// interface is the only place where the generator writes an FFI signature for a
+// custom type by hand, so it is the only place where the FFI type label can
+// disagree with the converter.
+[Custom]
+typedef string CustomId;
+
+callback interface CustomIdTransformer {
+    // `id` puts the custom type in argument position, and the return value puts
+    // it in return position. The generator lowers both to `RustBuffer`.
+    CustomId transform(CustomId id);
+};
+
 namespace custom_types {
     ZenEngineResponse get_zen_engine_response();
     ZenEngineResponse return_zen_engine_response(ZenEngineResponse response);
     APIResult roundtrip_api_result(APIResult value);
+    CustomId roundtrip_custom_id_through_callback(CustomIdTransformer transformer, CustomId id);
 };
--- a/fixtures/custom_types/src/lib.rs
+++ b/fixtures/custom_types/src/lib.rs
@@ -66,4 +66,35 @@ pub fn roundtrip_api_result(value: APIResult) -> APIResult {
     value
 }
 
+// Regression coverage for custom types in callback interface signature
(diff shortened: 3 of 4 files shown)


fixtures/custom_types/test/custom_types_test.dart
```diff
@@ -44,4 +56,21 @@ void main() {
...
+    final result = roundtripCustomIdThroughCallback(
+      transformer: transformer,
+      id: 'id-abc123',
+    );
+
+    // Rust passed the argument to Dart without a change.
+    expect(transformer.received, equals(<String>['id-abc123']));
+    // Dart returned a value, and Rust passed it back.
+    expect(result, equals('ID-ABC123'));
+  });
```
1egoman, line 75: Important callout: a callback interface is the only place that reaches the changed code. Just calling a top-level function would lower through `ffi_native_type_label`, which receives a `FfiType` that UniFFI has already resolved, so it never sees a custom type. A test that only calls a function therefore passes with or without the fix and isn't sufficient on its own.

1egoman:
@chavic For reference, the two ci failures on here seem to be the ones which we discussed on #150 (the [lint failure](https://github.com/Uniffi-Dart/uniffi-dart/pull/150#discussion_r3797663152), and [the flaky `dart_async` test](https://github.com/Uniffi-Dart/uniffi-dart/pull/150#issuecomment-5241530839) which I will leave unaddressed per https://github.com/Uniffi-Dart/uniffi-dart/issues/139)

Would you be willing to review this in its current state, or would you like to see the linting issue get resolved first? I think other than the linting issue this should be ready for review from my end.

1egoman:
Rebased on top of latest `main` to fix the lint errors.

chavic (approved):
Nice fix, thank you!

chavic:
The one test failure is the flaky `dart_async` timing test, which #157 has now fixed on `main`. This branch also conflicts with `main` in `src/gen/oracle.rs` after #150. One rebase on `main` corrects both. Everything else is green.

1egoman:
Thanks for the approval! @chavic Just rebased and all tests pass for me locally, let me know if there's anything else that you need from be before this can be merged.

1egoman:
@chavic Sorry for the ping, but anything else needed from me before this can be merged?

chavic:
@1egoman no worries, the pr seems to have fallen off my raider cause i was focusing on some other fixes... thanks for the good work and patience

PR #156, opened 2026-08-10
Move object converters into a sibling FfiConverter class

### LINKED ISSUE #155: Converter statics on object classes collide with Rust method names
## Summary

The generator puts the converter statics directly on the object class - `lift`, `lower`, `read`, `write` and `allocationSize`. A Rust method with one of these names then collides with the static of the same name. Since Dart does not permit a static member and an instance member with the same name in one class, the generated file does not compile.

`write` is the name most likely to collide (and what caused me to open this issue), because it is a common method name for any type that accepts bytes.

## Reproduction

This UDL is sufficient:

```webidl
interface Recorder {
    constructor();
    void write(bytes chunk);
    u64 size();
};

namespace example {};
```

## Actual generated Dart

One class holds both members:

```dart
class Recorder implements RecorderInterface {
  // ...
  static Pointer<Void> lower(Recorder value) { /* ... */ }
  static int allocationSize(Recorder value) { /* ... */ }
  static LiftRetVal<Recorder> read(Uint8List buf) { /* ... */ }
  static int write(Recorder value, Uint8List buf) { /* ... */ }   // converter

  void write({required Uint8List chunk}) { /* ... */ }            // Rust method
}
```

Dart reports:

```text
error - Class 'Recorder' can't define static member 'write' and have instance
        member 'Recorder.write' with the same name. Try renaming the member to
        a name that doesn't conflict. - conflicting_static_and_instance
```

(The same collision will occur for a Rust method named `lift`, `lower`, `read` or `allocationSize`).

## Proposed solution

I would think the easiest way to solve this is to put all the converter logic into its own class. From what I can tell, Records, enums and callback interfaces already follow this pattern, and the generator already names such a class `FfiConverter<Name>`. Objects should follow the same pattern:

```dart
class Recorder implements RecorderInterface {
  void write({required Uint8List chunk}) { /* ... */ }
}

class FfiConverterRecorder {
  static Recorder lift(Pointer<Void> ptr) { /* ... */ }
  static Pointer<Void> lower(Recorder value) { /* ... */ }
  static int allocationSize(Recorder value) { /* ... */ }
  static LiftRetVal<Recorder> read(Uint8List buf) { /* ... */ }
  static int write(Recorder value, Uint8List buf) { /* ... */ }
}
```

The name of a Rust method then cannot reach the converter statics.

1egoman:
FIxes #155 

## Summary

`ffi_converter_name` returned the canonical name of the object for `ObjectImpl::Struct` and `ObjectImpl::Trait`. The converter statics therefore landed on the object class itself, where a Rust method of the same name collided with them.

The function now returns `FfiConverter<CanonicalName>` for both variants, and `generate_object` and `generate_trait_object` emit that class as a sibling of the object. Records, enums and callback interfaces already follow this pattern.

## Effect on generated code

```diff
 class Recorder implements RecorderInterface {
-  static Pointer<Void> lower(Recorder value) { /* ... */ }
-  static int allocationSize(Recorder value) { /* ... */ }
-  static LiftRetVal<Recorder> read(Uint8List buf) { /* ... */ }
-  static int write(Recorder value, Uint8List buf) { /* ... */ }
-
   void write({required Uint8List chunk}) { /* ... */ }
 }
+
+class FfiConverterRecorder {
+  static Recorder lift(Pointer<Void> ptr) { /* ... */ }
+  static Pointer<Void> lower(Recorder value) { /* ... */ }
+  static int allocationSize(Recorder value) { /* ... */ }
+  static LiftRetVal<Recorder> read(Uint8List buf) { /* ... */ }
+  static int write(Recorder value, Uint8List buf) { /* ... */ }
+}
```

## Tests

I opted to not add any explicit tests for this since the existing test suite covers all these paths quite well - it's exercising all these existing internal bindgen method calls. If you'd like though, I could add something which goes through the whole call cycle with a method named `write()` on an object just to validate this 100% and protect against a regression.

### DIFF
--- a/src/gen/objects.rs
+++ b/src/gen/objects.rs
@@ -42,9 +42,14 @@ impl CodeType for ObjectCodeType {
 
     fn ffi_converter_name(&self) -> String {
         match self.imp {
-            ObjectImpl::Struct => self.canonical_name().to_string(), // Objects will use factory methods
+            // The converter lives in its own class rather than on the object itself, so
+            // that a Rust method named `read`/`write`/`lower` cannot collide with the
+            // converter's statics (Dart rejects a static and an instance member sharing
+            // a name). Matches how records, enums and callback interfaces are emitted.
+            ObjectImpl::Struct | ObjectImpl::Trait => {
+                format!("FfiConverter{}", self.canonical_name())
+            }
             ObjectImpl::CallbackTrait => format!("FfiConverterCallbackInterface{}", self.id),
-            ObjectImpl::Trait => self.canonical_name().to_string(),
         }
     }
 }
@@ -100,6 +105,7 @@ pub fn generate_object(obj: &Object, type_helper: &dyn TypeHelperRenderer) -> da
     }
 
     let cls_name = &DartCodeOracle::class_name(obj.name());
+    let ffi_converter_name = &obj.as_codetype().ffi_converter_name();
     let interface_name = DartCodeOracle::object_interface_name(type_helper.get_ci(), obj);
     let interface_definition = generate_object_interface(obj, &interface_name, type_helper);
     let finalizer_cls_name = &format!("{cls_name}Finalizer");
@@ -196,7 +202,7 @@ pub fn generate_object(obj: &Object, type_helper: &dyn TypeHelperRenderer) -> da
             class $(&error_handler_name) extends UniffiRustCallStatusErrorHandler {
                 @override
                 Exception lift(RustBuffer errorBuf) {
-                    return $(cls_name).read(errorBuf.asUint8List()).value;
+                    return $(ffi_converter_name).read(errorBuf.asUint8List()).value;
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/gen/objects.rs
```diff
@@ -692,6 +705,16 @@ fn generate_trait_object(obj: &Object, type_helper: &dyn TypeHelperRenderer) ->
...
 
+            void dispose();
+
+            $(for method in abstract_methods => $method)
+        }
+
+        class $ffi_converter_name {
+            static $cls_name lift(Pointer<Void> ptr) {
+                return $cls_name.lift(ptr);
+            }
```
1egoman, line 716: Important callout that this changes the generated API - `Recorder.lower(x)` becomes `FfiConverterRecorder.lower(x)`. Code that a consumer writes by hand I wouldn't think normally call these statics, because they belong to the FFI layer. But if any code does, it would need to change.

src/gen/objects.rs
```diff
@@ -196,7 +202,7 @@ pub fn generate_object(obj: &Object, type_helper: &dyn TypeHelperRenderer) -> da
             class $(&error_handler_name) extends UniffiRustCallStatusErrorHandler {
                 @override
                 Exception lift(RustBuffer errorBuf) {
-                    return $(cls_name).read(errorBuf.asUint8List()).value;
+                    return $(ffi_converter_name).read(errorBuf.asUint8List()).value;
                 }
```
1egoman, line 206: Note that the error handler now points at the converter - an `[Error] interface` emits an error handler class, and that handler called `Recorder.read`. It now calls `FfiConverterRecorder.read`.

1egoman:
@chavic For reference, the two test failures on here seem to be the ones which we discussed on #150 (the [lint failure](https://github.com/Uniffi-Dart/uniffi-dart/pull/150#discussion_r3797663152), and [the flaky `dart_async` test](https://github.com/Uniffi-Dart/uniffi-dart/pull/150#issuecomment-5241530839) which I will leave unaddressed per https://github.com/Uniffi-Dart/uniffi-dart/issues/139)

Would you be willing to review this in its current state, or would you like to see the linting issue get resolved first? I think other than the linting issue this should be ready for review from my end.

1egoman:
Rebased on top of latest `main` to fix the lint errors.

chavic (approved):
Nice fix, thank you. On your API callout: the rename is safe to take. The generator emits these calls from one source, so no stale call site remains, and the hand-written Dart in payjoin-ffi and bdk-dart calls none of these statics.

Please add a small fixture with an object that has a method named `read` or `lower`. That is the collision this change corrects, and no test guards it yet.

PR #160, opened 2026-08-28
Use monotonic clock and inclusive lower bounds in dart_async test

chavic:
## What
- targeting the flaky async time tests
- `measureTime` now uses a monotonic `Stopwatch` instead of `DateTime.now()`.
- The three lower-bound assertions that used a strict `>` now use `>=`, like the rest of the file.

## Why

`DateTime.now()` is the wall clock. NTP can slew or step it on a CI runner during a measurement, so the interval it reports can be shorter than the time that actually elapsed. Rust's `thread::sleep` never returns early, so the lower-bound failure on #149 (`sleep(200)` measured as `<= 200ms`) can only come from the clock, not from the code under test.

A sleep of exactly N ms also truncates to N milliseconds, which fails a strict bound. Two of the three strict bounds sit on a single sleep, so this is latent on any fast runner.

Refs #139.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fixtures/dart_async/test/futures_test.dart
+++ b/fixtures/dart_async/test/futures_test.dart
@@ -2,10 +2,10 @@ import 'package:test/test.dart';
 import '../dart_async.dart';
 
 Future<Duration> measureTime(Future<void> Function() action) async {
-  final start = DateTime.now();
+  final stopwatch = Stopwatch()..start();
   await action();
-  final end = DateTime.now();
-  return end.difference(start);
+  stopwatch.stop();
+  return stopwatch.elapsed;
 }
 
 // The DELAY-based timing assertions below check a LOWER bound only: that an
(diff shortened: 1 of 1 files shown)


fixtures/dart_async/test/futures_test.dart
```diff
@@ -1,11 +1,16 @@
 import 'package:test/test.dart';
 import '../dart_async.dart';
 
+// Uses a monotonic Stopwatch, not DateTime.now(). DateTime.now() is the wall
+// clock, which NTP can slew or step during a measurement on a CI runner, so an
+// interval it reports can be shorter than the time that actually elapsed. That
+// makes a lower-bound assertion fail even though the sleep ran for its full
+// duration (#139).
```
spacebear21: This kind of comment belongs in the commit message, not a code comment (common frustrating pattern from the robots to overly explain things in comments...)

spacebear21 (approved):
utACK

fixtures/dart_async/test/futures_test.dart
chavic: Agreed, nice catch, I'll clean it up

chavic: Agreed, nice catch, I'll clean it up

Issue #161, opened 2026-09-11
Rust worker-thread callbacks abort the Dart VM


chavic:
I found a process abort when Rust invokes a generated Dart callback from a worker thread. The callback accepts `Send + Sync`, but the generated entry point uses `Pointer.fromFunction`, which cannot be invoked outside the creating isolate's mutator thread.

The Rust trigger is:

```rust
use std::sync::Arc;

#[uniffi::export(with_foreign)]
pub trait Sink: Send + Sync {
    fn bytes(&self, value: Vec<u8>) -> u64;
}

#[uniffi::export]
pub fn call_bytes_thread(sink: Arc<dyn Sink>) -> u64 {
    std::thread::spawn(move || sink.bytes(vec![42])).join().unwrap()
}
```

With a Dart implementation that just returns the byte length, the subprocess exits 134 with `Cannot invoke native callback outside an isolate`.

We need a thread/isolate dispatch design for these callbacks, including clone/free and foreign-future cleanup. Simply replacing the entry point with an asynchronous listener won't cover synchronous callbacks: Rust expects the output and status pointers to be filled before the callback returns, and those pointers cannot safely escape that call.

Please keep a subprocess regression for the worker-thread case so a VM abort does not take down the whole test runner. Related to #65.

This is a separate follow-up to the audit around #149. The affected code already exists on its base (`1ded89b`), so fixing this does not depend on merging that PR. Runtime probes were run against `50afc12` on Linux x86_64 with Dart 3.13.0; inheritance was checked against the base source.

Relevant base code: [src/gen/callback_interface.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/callback_interface.rs#L462).

chavic:
I've put a synchronous relay prototype in #186 so we can test the approach before changing the Dart API.

The unchanged bindings reproduce the worker-thread abort. The relay passes the actual generated-UniFFI call on Linux x64, including concurrent workers, nested calls, errors and clone/free. It also keeps the original Dart callback state, which the test checks separately from the return value.

I haven't found anyone reporting this downstream so far. This is a reproduced limitation, not a confirmed downstream incident.

[...]

PR #173, opened 2026-09-11
Complete async callbacks when result conversion fails

### LINKED ISSUE #163: Async callback result-conversion failures suppress native completion
An async callback can finish in Dart but never notify its native caller if converting the result throws.

```rust
#[uniffi::export(with_foreign)]
#[async_trait::async_trait]
pub trait AsyncSmall: Send + Sync {
    async fn small(&self) -> u8;
}
```

A Dart implementation returning `42` completes normally. Returning `256` throws during the generated `u8` lowering, but no error completion is sent.

The success path removes the callback handle and sets `effectiveState.cancelled = true` before allocating/lowering the result. If that work throws, the catch block sees `cancelled` and returns immediately.

I tested the generated trampoline with a real native completion pointer: `42` completed, while `256` hit the 200 ms test timeout. The probe did not leave a permanently waiting Rust future running.

We need to distinguish cancellation from completion in progress and ensure a conversion failure produces exactly one error completion. Coverage should include successful results, invalid results and cancellation races. This is a deterministic error-path defect, separate from the timing investigation in #139. Related to #65.

Relevant base code: [src/gen/callback_interface.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/callback_interface.rs#L398).

chavic:
Following up on #163. If an async Dart callback returns `256` for a Rust `u8`, result conversion throws after the callback has already been marked cancelled. The catch branch then returns without notifying Rust, leaving the caller waiting.

This separates completion from cancellation and only marks completion once the result or error is ready to pass back. A failure while lowering the result can therefore complete with an error, and a cancelled callback still stays cancelled.

I've tested this through a real Rust future and directly through the generated callback. The async suite passed all 36 Dart tests, including eight new cases covering invalid results, declared and unexpected errors, cancellation, and cleanup during completion. Restoring the old ordering makes the new invalid-result test time out, so it catches the original bug. Formatting and diff checks passed too.

Callback threading remains in #161; failures while serializing the error itself and ownership rollback inside complex converters remain separate work.

Fixes #163.


chavic:
@DenisovAV could you review this one? It makes sure async callbacks still complete when converting their result fails.

DenisovAV (commented):
Fixes the #163 case: reverting it makes two of the new cases fail, one hangs until the timeout.

Two paths can still leave the Rust future waiting forever:
1. Lowering the error throws: a declared error field that fails to lower, or an `e.toString()` that throws. A probe confirmed no completion and a hang. The description scopes this out, but #177 and #187 add exactly these throws (a negative `Duration` or unrepresentable timestamp inside an error). Setting `CALL_UNEXPECTED_ERROR` first and falling back to a plain message if lowering fails would close it.
2. The handle lookup and arg lifts run before the `try`, so a stale handle or a failing lift (#187 makes an out-of-range timestamp arg one) escapes without completing. Moving them into the closure's `try` doesn't change timing, the async body runs synchronously until the first `await`.

So I'd merge this before or together with #177 and #187.

Smaller:
- A cancelled future is never completed, which leaks one oneshot `Arc` per cancellation on the Rust side (Kotlin still completes after cancellation). The new tests assert `completions == 0`, which locks that in.
- The "releases its future handle" check can't fail: the test's completion callback calls `free`, so removing all three `maybeRemove` calls still passes 8/8.
- `final free = ...` in `callback_completion_test.dart` is unused.

PR #176, opened 2026-09-11
Escape dollar signs in Dart string defaults

### LINKED ISSUE #168: Escape dollar signs in generated Dart string defaults
A literal dollar sign in a Rust string default becomes Dart interpolation and can make the generated library fail analysis.

```rust
#[derive(uniffi::Record)]
pub struct Price {
    #[uniffi(default = "$amount")]
    pub text: String,
}
```

The generator emits a default equivalent to `Price({this.text = '$amount'})`. Dart treats `amount` as a variable, producing undefined-identifier and nonconstant-default errors instead of preserving the literal string.

`escape_dart_string` handles other escapes but leaves `$` unchanged. Escape it in generated literals and add coverage for `$amount`, `${amount}`, quotes and backslashes, checking that the default value roundtrips unchanged. This is a remaining correctness case after default support in #142.

Relevant base code: [src/gen/primitives/mod.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/primitives/mod.rs#L23).

chavic:
String defaults containing `$amount` or `${amount}` currently become Dart interpolation. That can stop the generated bindings from compiling instead of preserving the literal value.

This escapes dollar signs alongside the existing backslash and quote escaping. I've added fixture coverage for defaults from both UDL and proc-macro records, including a Rust roundtrip with quotes, backslashes, and dollar signs.

I've checked the generated bindings with all eight fixture tests passing. Removing the escaping from the tested defaults makes Dart report undefined `amount` identifiers, so the new cases catch the original failure. Formatting, diff checks, and all 12 CI checks pass too.

Fixes #168.


chavic:
@DenisovAV could you review this one? It fixes dollar signs in generated Dart string defaults.

DenisovAV (commented):
LGTM. Backslash before `$` is the right order, and both literal paths go through `escape_dart_string`, so record-field and argument defaults are covered. Reverting the change breaks the fixture's compile, so the test guards it.

Optional: a test with `$` in a function argument default (`#[uniffi::export(default(x = "$y"))]`). Covered by construction, just not exercised.

PR #177, opened 2026-09-11
Reject negative durations before serialization

### LINKED ISSUE #170: Reject negative Dart Duration values before lowering to Rust
The generated duration converter accepts negative Dart values even though Rust's `std::time::Duration` is nonnegative.

This direct converter call succeeds and reports 12 bytes written:

```dart
FfiConverterDuration.write(
  Duration(microseconds: -1),
  Uint8List(12),
);
```

The seconds/nanoseconds are serialized into unsigned fields, so the negative input is wrapped instead of rejected. This is an input-validation gap in the existing duration support from #26.

Reject negative durations before writing any native argument, with a clear Dart error. Cover negative subsecond and whole-second values, zero and positive values, including use inside compound values. The observed failure is converter acceptance/serialization; I did not rely on passing the invalid value into Rust.

Relevant base code: [src/gen/primitives/duration.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/primitives/duration.rs#L34).

chavic:
Negative Dart durations currently get written into unsigned seconds and nanoseconds instead of being rejected.

This throws `ArgumentError` when sizing or writing a negative duration. The size check also catches negative values nested in records, optionals, and lists before those values are serialized. Direct writes reject the value before changing the destination buffer.

I've added coverage for negative subsecond and whole-second values, unchanged storage after rejection, zero, and positive scalar and compound roundtrips. All nine fixture tests pass on Rust 1.85. Removing the guard makes the direct serialization test accept the negative value again, so it catches the original bug. Formatting, diff checks, and all 12 CI checks pass too.

Fixes #170.


chavic:
@DenisovAV could you review this one? It rejects negative durations before they reach unsigned serialization.

DenisovAV (commented):
Write side looks right: the check in `allocationSize` rejects a negative value nested in a record/optional/list before anything is written, and reverting the fix fails the three negative-value tests.

The read side still wraps though. `read` has no range check, so `Duration::MAX` comes back as -1µs and anything above ~9.22e12 s turns negative (checked by running the generated `read`). With this PR that value then can't go back to Rust, it hits the new `ArgumentError`, and `Duration::MAX` is a common "no timeout" value. A range check in `read` like #187 does for timestamps would fix it, here or in a follow-up.

Minor: the new throw also reaches async callback paths, so best merged with or after #173.

PR #178, opened 2026-09-12
Avoid shadowing status arguments in generated calls

### LINKED ISSUE #169: Generated call-status variables shadow user parameters named status
The generated `(status)` closure parameter shadows an exported argument with the same name. Lowering the user argument inside that closure then reads the `Pointer<RustCallStatus>` instead of the supplied value.

The base generator already emits this shape for ordinary synchronous functions:

```dart
int echoStatus({required int status}) {
  return rustCallWithLifter(
    (status) => nativeEcho(FfiConverterUInt32.lower(status), status),
    FfiConverterUInt32.lift,
  );
}
```

This illustrates the inherited collision for a Rust `fn echo_status(status: u32) -> u32`. The executed analyzer probe used the borrowed-byte form on #149 and reported that `Pointer<RustCallStatus>` cannot be passed as `Uint8List`. I checked the ordinary argument-lowering template on the base; I did not rerun that primitive example as a separate runtime probe.

We need generated local names that cannot shadow user parameters, applied consistently across functions, constructors and methods. Add generator/analyzer coverage using `status` as a primitive parameter so the regression test does not rely on the new borrowed-byte support.

Relevant base code: [src/gen/functions.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/functions.rs#L70), [src/gen/objects.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/objects.rs#L182).

chavic:
An argument named `status` currently gets shadowed by the generated call-status variable, so valid APIs produce Dart that won't compile. The same collision also affects synchronous callback dispatch.

This checks the arguments' Dart names before choosing the internal name, including cases such as `status_1` becoming `status1`. It covers functions, constructors, object and trait methods, and callbacks without changing their public parameters.

I've tested both constructor forms, void and value returns, declared errors, and calls in both directions through a foreign trait. All nine proc-macro fixture tests pass on Rust 1.85.1, including five new regressions. Restoring the old names in the generated bindings brings back the compiler failures. Formatting, diff checks, and all 12 CI checks pass too, including BDK and Payjoin downstream tests.

Fixes #169.


chavic:
@DenisovAV could you review this one? It keeps generated call-status variables from shadowing user parameters.

DenisovAV (commented):
Complete for `status`: reverting it gives 16 compile errors across functions, constructors, methods, trait impls and sync callbacks.

Other generated names have the same problem, and one is silent. A sync callback `two_ints(a, arg0)` generates `final arg0 = lift(a); final arg1 = lift(arg0);`, so the method gets the wrong value (a probe returned 11 instead of 12). `out_return`, `uniffi_handle`, `obj`, `result`, and in async callbacks `state`, `handle`, `callback`, fail to compile. `var_name` never produces a leading underscore, so `_uniffi`-prefixed names for generated locals and params would rule all of this out. Fine as a follow-up.

Pre-existing, but in code this touches: in the sync callback `catch`, if lowering the declared error throws (or `e.toString()` does), the exception escapes a void `fromFunction` callback, Dart drops it, and Rust sees `Success` with a default return value. Might be worth an issue.

PR #185, opened 2026-09-12
Make CI failures and test coverage explicit

### LINKED ISSUE #184: Nightly CI masks a rustix compilation failure before tests run
The nightly test job is reporting green even though the workspace never reaches its tests. The test step uses `continue-on-error`, and rustix 0.37.28 fails compilation with errors about reserved `rustc_layout_scalar_valid_range_*` attributes.

I've checked this on main at e2dd2bb, so it predates the ownership fixes: [main nightly log](https://github.com/Uniffi-Dart/uniffi-dart/actions/runs/34608737385/job/103293577938). The same failure occurs in subsequent draft runs. Rust 1.85 and stable complete their test suites.

The dependency comes from the streams_ext fixture through smol 1.3.0 and async-io 1.13.0. `cargo tree -i rustix@0.37.28 --workspace` shows that path. The failing CI command is `cargo nextest run --all --all-features` on nightly.

We should fix that dependency path and make it clear when the allowed-to-fail nightly step has not actually run tests. The green job status currently hides the gap.

chavic:
Nightly can currently report green without reaching the tests, and the Nix tool check can succeed after an earlier command fails. Clippy also only selects the root package, while the BDK result includes skipped integration tests.

This makes nightly test and lint failures fail their jobs, keeps the other matrix jobs running, and makes the Nix shell stop on the first failed tool. The stream fixture uses async-io directly for its timer and executor, removing the old rustix dependency that breaks nightly. Rust 1.85 and UniFFI 0.31.2 stay in place.

Clippy now checks the whole workspace. I've cleaned up the fixture warnings this exposed, with narrow allowances for intentional fixture behavior and UniFFI's generated metadata constants. The Nix summary states that its flake check covers formatting.

BDK now reports passed, failed and skipped test counts in the job summary. A manual downstream run accepts Electrum and Esplora testnet endpoints, enables integration tests, and fails on skipped tests or an incomplete report. Routine PR runs remain independent of live testnet servers.

I've tested workspace Clippy on Rust 1.85.1 and recent nightly, all eight Rust and seven Dart stream tests on recent nightly, eight result-reporting tests, and real Dart reports with passing, skipped and failed tests. The strict report check rejects skipped runs even when Dart exits successfully. Injected Nix tool failures now return failure too.

CI passed all 41 workspace tests with zero skips on Rust 1.85, stable and nightly. All 12 checks passed, including lints, Nix, BDK and Payjoin. Routine BDK reported 25 passed and five integration tests skipped.

The [manual BDK integration run](https://github.com/chavic/uniffi-dart/actions/runs/34698679387) passed all 30 tests with zero skips, using the Blockstream testnet Electrum and Esplora servers.

Fixes #184.


reez (approved):
ACK 32497edfc08507bb718ee6add309920a5fc41999

nit, doesn't need to block: is the comment at `.github/scripts/report-dart-tests.py:16` accurate? (I ask because I think Dart's JSON reporter only sends testDone once per test, and errors that happen after a test finishes come as separate error events, so the exit status is still correct because of `done.success`, but the pass count can be off?

PR #186, opened 2026-09-13
Generate opt-in dispatch for synchronous callbacks

chavic:
Rust worker-thread callbacks can abort the Dart VM. This draft now includes an opt-in generated native bridge, beyond the handwritten relay used to investigate it. I still haven't found a downstream report of the original crash.

With `callback_dispatch = true`, the generator emits the Dart wrappers and a standalone Rust companion crate. Native calls run on workers while the owning Dart isolate services callback requests. Retained callbacks use a listener between calls. Each isolate has its own endpoint and copied vtables, and native callback handles route back to the correct local registry. Nested calls share the owner's queue. The Dart calls stay synchronous and keep the original callback objects.

The companion uses only Rust's standard library and is bundled as a second Native Asset. It calls the original library through function pointers resolved by Dart, so it doesn't link another copy of the Rust application. The integration guide includes the build command, asset registration and shutdown API.

I've tested 12 generated-bridge scenarios in both JIT and compiled AOT on Linux x64 with Dart 3.13.0, Rust 1.85.1 and UniFFI 0.31.2. These cover worker/join and parallel calls, declared errors, nested calls through an outer handle, retained callbacks, 100 repeated calls, 50 retained clone/free cycles and two concurrent Dart isolates. Removing callback-registration rollback makes the failed-lowering shutdown test fail. A state-mutation control fails even when the callback returns the expected scalar. Native shutdown tests reject close during an active call or a pending listener notification. The generator and native companion pass Rust 1.85.1 Clippy with warnings denied. All 45 existing Dart callback and async tests pass with dispatch disabled. The earlier relay and wrong-thread abort control remain in the fixture.

This first generated version is deliberately limited to synchronous function APIs on 64-bit targets. Generation rejects async APIs/callbacks, Rust object types, returned or nested callback values, external component types and the direct callback wire types with existing converter mismatches. Forced isolate termination and cancellation still need work. Native calls move to fresh workers, so libraries with thread-affinity requirements must not opt in. Applications must release retained callbacks and close the dispatcher before ending its isolate.

The feature is off by default. The dedicated CI runs both the older controls and the generated bridge in JIT/AOT. Fresh device results are needed; the earlier macOS/iOS/Android reports apply to the handwritten prototype.

```sh
python3 fixtures/callback-thread-relay/probe/run_generated.py --dart /path/to/dart
python3 fixtures/callback-thread-relay/probe/run_generated.py --dart /path/to/dart --aot
```

See `docs/callback-dispatch.md` for the supported scope and packaging contract. Related to #161; this does not close the remaining unsupported paths.


chavic:
@DenisovAV could you review this one? The generated callback bridge and its supported scope need a look. JIT and AOT probes pass on Linux x64; more device results would help too.

DenisovAV:
macOS arm64 results for c76728a, since you asked for more devices.

Apple M4 Pro, macOS 26.5.1 (25F80), Dart 3.13.3, UniFFI 0.31.2, rustc/clippy 1.88.0 (1.85.1 wasn't installed, so I used the runners' `--toolchain` flag).

| Runner | JIT | AOT |
|---|---|---|
| `run_generated.py`: 12 cases + state and rollback controls, companion tests + clippy | 3/3 | 3/3 |
| `run.py`: handwritten relay, 11 cases + baseline + mutation | 3/3 | 3/3 |

- 4,800 AOT runs (each case 200× serially and 200× eight at a time) plus 180 JIT runs, no failures.
- #161 reproduces on arm64: with dispatch off, `baseline-thread` aborts in JIT and AOT with `Cannot invoke native callback outside an isolate`.
- The suite catches the bridge being removed: patching the companion so `dispatch()` calls back on the calling thread makes 11 of 12 cases abort, only `rollback` (no callback) passes.
- JIT ~0.7–1.0 s per case warm, AOT 0.02–0.16 s.

The disk filled up mid-session, so after the first full run the later ones reused its build and skipped only `cargo build`; everything else ran as is.

The runners already handle `.dylib`, so `macos-latest` can just be added to `callback-thread-probe.yml`. Windows is still untested.

DenisovAV (commented):
With `callback_dispatch` off the output is token-identical to before, so existing users aren't affected. Notes on the opt-in path:

Hangs and crashes not covered by the docs yet:
- Thread-local drop deadlock. Every call runs on a fresh worker, and `run` stops pumping once the result arrives, then blocks in `join()`. If Rust keeps a callback in a `thread_local!`, the worker's exit drops it, the `uniffi_free` proxy queues a job and waits, and nobody drains the queue. Reproduced with the runtime template in a small Rust harness. Not joining, or a persistent worker, would avoid it.
- A callback invoked on the owner's own thread outside a dispatched call (a hand-written `@Native` export, or platform code on Flutter's main thread) queues work to itself and blocks forever. Before this it was a loud abort.
- `Isolate.run` / `compute` / hot restart with a retained callback crash the process later with `Callback invoked after it has been deleted` (3/3 in JIT and AOT). It's documented as unsupported, but `Isolate.run` is the usual way to offload FFI, so a fail-fast check or a prominent warning would help.
- The listener keeps the isolate alive after any binding call (`keepIsolateAlive` defaults to true), so a CLI that calls one function never exits without `closeCallbackDispatcher()`. Keeping it alive only while routes are live would fix that.

Robustness:
- Nothing checks that the separately built companion matches the bindings. A stale one means mismatched `dispatch_*` signatures and silent memory corruption. A hash exported by the companion and checked in the `_UniffiDispatch` constructor would turn it into a clear error.
- `ObjectImpl::CallbackTrait` doesn't exist in uniffi_meta 0.32; `Type::Object { imp, .. } if imp.has_callback_interface()` works on both 0.31.2 and 0.32.1. The probe also pins 1.85.1, which clashes with the 1.88 MSRV on the 0.32 branch. I can take both when I rebase #149.
- Rejecting 32-bit is right, but `flutter build apk` builds armeabi-v7a by default, so the companion build fails (or every call throws). Worth documenting `--target-platform`.

Reviewability: roughly 1,800 of the 3,132 lines are the handwritten prototype and recorded reports. `DESIGN.md` and the fixture README still describe generator support as future work, `VALIDATION.md` references commit `816ea89` and a `review/...` path that aren't in the repo, and `results/linux-x64.json` isn't referenced anywhere. Moving the prototype and reports to #161 or a research branch would make the generator change much easier to review.

Scope-wise: validation rejects Rust objects and async, so bdk, payjoin and qdrant-edge can't opt in yet, and #161 stays open for them, as you say.

PR #187, opened 2026-09-13
Support timestamps as Dart DateTime

### LINKED ISSUE #25: Timestamp Type support

chavic:
Exporting Rust `SystemTime` currently panics during Dart generation. This maps timestamps to `DateTime` and activates the `time-types` fixture in the workspace.

Returned values use UTC, and local DateTime inputs preserve the same instant. Nanoseconds below microsecond precision are truncated toward the Unix epoch. The converter checks the DateTime range before integer multiplication, validates nanoseconds and buffer lengths, and uses RustBuffer signatures for timestamp callbacks.

There is one upstream format limitation: UniFFI 0.31.2 stores the sign in whole seconds, so timestamps strictly between one second before the epoch and the epoch lose their negative sign. A Rust-only serialization probe confirmed this. Dart inputs in that interval throw rather than silently change their instant; Rust outputs have already lost the sign before Dart receives them. The README documents this alongside the precision behavior.

Validation on Linux x64 with Rust 1.85.1 and Dart 3.13.0:

- The baseline reproduces `Renderable for Type::Timestamp` during generation.
- All 15 Dart tests pass through the Cargo fixture runner. They cover functions, constructors/methods, callbacks, async returns, records, optionals, lists/maps, UTC conversion, precision boundaries, malformed buffers and range errors.
- The generator's library test passes. Clippy passes for both the generator and time-types with all targets/features and warnings denied. Rust/Dart formatting and diff checks pass.

The converter follows the existing borrowed-buffer convention so the separate incoming-buffer ownership fix can release results at the FFI boundary.

Fixes #25.


chavic:
@DenisovAV could you review this one? It adds timestamp support using Dart DateTime.

DenisovAV (commented):
Converter matches UniFFI's sign-magnitude encoding, range checks run before the multiplication, and the tests are strong: I injected eight different bugs into the generated converter and all eight were caught.

A few things:
- A user type named `Timestamp` clashes with the built-in: a crate with a `Timestamp` record plus a `SystemTime` function generates two `class FfiConverterTimestamp`. Not a regression (`SystemTime` used to crash the generator), but it's a common name, and `Duration` probably has the same issue. More of an issue than a blocker.
- Worth validating in `allocationSize` too, like #177. Otherwise in a record like `{obj, at: timestamp}` the object's pointer is cloned before `write` throws.
- README: returned values are UTC and `DateTime ==` compares `isUtc`, so an echoed local value isn't `==` to the input. A line pointing to `isAtSameMomentAs` would help.

FYI for #149: this removes the `_ => todo!()` fallbacks and 0.32 adds `Type::Set`, so I'll add explicit arms when I rebase the 0.32 branch on top.

PR #190, opened 2026-09-16
Use fixed-width UniFFI handles on 32-bit targets

### LINKED ISSUE #171: Use the fixed-width UniFFI Handle ABI on 32-bit targets
The generator maps `FfiType::Handle` to `Pointer<Void>` in both native and Dart signatures. UniFFI 0.31.2 already defines `Handle` as `#[repr(transparent)] pub struct Handle(u64)`.

On a 32-bit target, a pointer is 32 bits while this wire value is still 64 bits. Those types are not ABI-compatible, even when a Rust-owned handle originated from a pointer. The mismatch predates the UniFFI 0.32 upgrade.

This finding is from source inspection, including the installed `uniffi_core-0.31.2/src/ffi/handle.rs`; I have not run a 32-bit reproduction. It should be treated as an ABI defect to validate, not a reported device crash.

Audit handle signatures and transport end to end across objects, futures and callbacks. Preserve the fixed-width ABI and keep actual pointer types distinct. Add a 32-bit compile/runtime check for handle creation, cloning, calls and disposal before claiming that target is covered.

Relevant base code: [src/gen/oracle.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/oracle.rs#L120), [src/gen/oracle.rs](https://github.com/Uniffi-Dart/uniffi-dart/blob/1ded89be15e2d18346df8701df2b9cfcae918316/src/gen/oracle.rs#L145).

chavic:
A Rust object method returning a declared error can crash Dart on ARM32 instead of throwing the generated exception. UniFFI uses a 64-bit handle, but we were declaring it as a pointer. That shifts the following arguments on 32-bit targets, including the call-status pointer. A successful getter can still work, which hid the problem in the first test.

This uses Uint64 at the native boundary and opaque int handles throughout objects, traits, callback converters and Rust futures. Clone/free calls, serialized handles, callback object returns and future continuation data use the same representation. Ordinary Dart object and Future APIs stay the same. Regenerate the component bindings and shared runtime together.

I've tested the generated bindings on Linux x64 and ARM32 under QEMU with Dart 3.13.0, Rust 1.85.1 and UniFFI 0.31.2. The regression covers typed object errors, object arguments and returns through callbacks, Rust-only and foreign traits, and a pending Rust future returning an object. It also passes with callback and continuation handles using nonzero upper bits and the sign bit.

Reverting only the failing method's declaration to Pointer<Void> makes the ARM32 process abort again at the typed error. The same mutation passes on x64. Native ABI recorders separately confirm the argument, return and callback layouts: six Uint64 controls pass on ARM32 while all five old pointer declarations disagree.

The new simple-interface regression and the existing callback, async, proc-macro, trait-interface and trait-method Dart suites all pass locally. The generator runtime test also passes. Rust 1.85.1 Clippy passes for the generator and fixture with all features, all targets and warnings denied.

The fixture README includes the standalone runner, cross-build instructions and validation limits. Normal fixture CI runs the new case on its host; the ARM32/QEMU matrix is a separate run. No physical mobile-device or ARM32 AOT result is claimed, and I haven't identified a downstream report of this crash.

Fixes #171.


reez:
Concept ACK

`FfiType::Handle` now renders as a Dart `int`, but `foreign_future_field_annotation` has no `Handle` arm, so async foreign trait methods that return an object would get a result field with no annotation. I tested this with an async `with_foreign` trait returning an object: it passes on main, but on this PR Dart fails to compile with `Field 'returnValue' requires exactly one annotation`. 

(Adding `FfiType::UInt64 | FfiType::Handle => Some(quote!(@Uint64()))` fixed it for me at least?)

